PR #8, opened 2020-06-29
Add UI kit to project

ramyabuva:
https://screenshot.googleplex.com/XLGwg9uOKTe.png


savsa (approved):

src/main/webapp/index.html
```diff
@@ -1,11 +1,263 @@
 <!DOCTYPE html>
```
wli, line 1: When importing large bundles like this, you should not make any changes to other source code. Keeping this import clean really helps set the boundary of "this is our code" vs "this is third-party imported code".

src/main/webapp/awesome-dashboard-ui-kit/assets/vendor/themify-icons/SVG/video-clapper.svg
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
```
wli, line 1: This is a red flag to me. I think this is another project via https://themify.me/themify-icons. There shouldn't be any need to commit these files to source code. You should be able to load these via CDN (https://www.jsdelivr.com/package/npm/themify-icons).

wli (changes requested):
High level feedback: while this UI kit seems tempting to use, ongoing maintenance may be quite difficult.

It looks like this distribution contains bundled files for other dependencies, including bootstrap, jquery, and others. There should be no reason to commit versions of these third-party libraries to our source code for several reasons:

1. It clutters up the code base.
2. They can easily get out of date. Especially for stuff like bootstrap and jquery, there are frequent updates that fix real bugs that should be getting pulled in.

At the bare minimum, I would request that you delete all sub-dependencies (the ones in `vendor/`) and pulling those in via CDNs.

Beyond that, as this theme uses bootstrap as a base, I would highly recommend seeing if you can simply use a lightweight bootstrap template instead to recreate this template. A few reasons:

1. It has passable UI, with a standard dependency model without needing to do this massive import.
2. Bootstrap is a relatively popular template, so the skills/knowledge you learn can be carried over to future projects in your career.
3. It doesn't look like Awesome Dashboard UI Kit is really maintained, as it has one "download button" and is using outdated versions of libraries.
4. If you run into any issues with bootstrap, there are tons of examples of how to get bootstrap to do something - as opposed to little-to-no examples of how to tweak Awesome UI kit.

Happy to chat quickly about this (tomorrow's pod sync perhaps) on the pros/cons of this more.

PR #11, opened 2020-06-30
Add UI template

ramyabuva:
https://screenshot.googleplex.com/MHtfNCwAeps.png

Template link: https://startbootstrap.com/themes/sb-admin-2/
CDNs used for all vendor .css and .js files


savsa (approved):

src/main/webapp/script.js
```diff
@@ -1,14 +0,0 @@
-// Copyright 2020 Google LLC
```
wli, line 1: i think you will still need this js and css file? Where will you put your own custom javascript or css?

triplehair (approved):

savsa (approved):

src/main/webapp/LICENSE
```diff
@@ -0,0 +1,21 @@
+The MIT License (MIT)
```
wli: This isn't quite right. By placing the MIT license here, you would be applying this MIT license (with copyright "Start Bootstrap LLC") to all files in this directory, which is supposed to be your code.

Instead, you should create a new directory (we can call it `src/main/webapp/vendor`), where you should place both this LICENSE file, and also the index.html file (since that was derived from their code). Future updates to index.html would happen in `vendor/`. For any other HTML templates you create on your own, you can place those in `src/main/webapp/` proper.

You can (hopefully) see more guidance here: https://groups.google.com/a/google.com/g/third-party-help/c/uVob1Bnqwv0/m/Gj8TdWVbCQAJ. If you can't, the short version is noted in the first paragraph of my comment here.

wli (changes requested):

src/main/webapp/index.html
```diff
@@ -1,11 +1,194 @@
...
 <html lang="en">
-  <head>
-    <meta charset="UTF-8">
-    <title>Our Capstone Project</title>
-    <link rel="stylesheet" href="style.css">
-    <script src="script.js"></script>
-  </head>
-  <body>
-  </body>
+
```
wli: please add this file to `.gitignore`. We shouldn't commit this file.

wli (changes requested):

src/main/webapp/vendor/index.html
```diff
@@ -0,0 +1,194 @@
+<!DOCTYPE html>
```
wli: this should be in `vendor/awesomedashboard2/index.html` or something. Same with the LICENSE.

wli (changes requested):

src/main/webapp/script.js
wli: This likely still needs to come back, but we can do it later.

wli (approved):
Approving and merging to unblock future dev work. Thanks for adding this, and handling the changes here!

PR #18, opened 2020-07-08
Create and integrate Planner servlet

ramyabuva:
https://screenshot.googleplex.com/JHLnc37Dktx.png


src/main/webapp/script.js
```diff
@@ -100,21 +100,84 @@ window.addEventListener('load', () => {
...
+        createSpinner(courseContainer);
+        const response = await fetch(
+            `/planner?selectedClasses=${selected.toString()}&semesters=${
+                document.getElementById('semesters').value}`);
+        courseList = await response.json();
+        createTable(courseList, courseContainer);
+      }
+
+      /**
+       * Creates table from a 2D array
```
savsa: Comment is wrong

src/test/java/com/google/collegeplanner/TestPlanner.java
```diff
@@ -0,0 +1,40 @@
...
+
+import com.google.collegeplanner.servlets.PlannerServlet;
+import java.io.PrintWriter;
+import java.io.StringWriter;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import org.json.simple.JSONArray;
+import org.json.simple.JSONObject;
+import org.json.simple.parser.JSONParser;
```
savsa: Unused imports

src/test/java/com/google/collegeplanner/TestPlanner.java
```diff
@@ -0,0 +1,40 @@
...
+import org.junit.Assert;
+import org.junit.Before;
```
savsa: Also unused

src/test/java/com/google/collegeplanner/TestPlanner.java
```diff
@@ -0,0 +1,40 @@
...
+    HttpServletResponse mockedResponse = mock(HttpServletResponse.class);
+    StringWriter stringWriter = new StringWriter();
+    PrintWriter writer = new PrintWriter(stringWriter);
+    when(mockedResponse.getWriter()).thenReturn(writer);
+
+    PlannerServlet servlet = new PlannerServlet();
+    servlet.doGet(mockedRequest, mockedResponse);
+    writer.flush();
+    // Check whether the string output is correct
+    System.out.println(stringWriter.toString());
```
savsa: Take the print statement out

src/main/webapp/script.js
```diff
@@ -100,21 +100,84 @@ window.addEventListener('load', () => {
...
-      useDetailPopup: true,
-      disableDblClick: true,
-      disableClick: true,
-      isReadOnly: true,
-      scheduleView: ['time'],
-      taskView: false,
-      week: {
-        workweek: true,
-      },
+  if (document.head.id == 'scheduler') {
```
savsa, line 103: I feel like this is kinda strange. I think it would be better to have separate js files for each html page and extract the reused functions into its own js file. Then you could just import that file at the top of this file and use the functions. I've just never seen js that depends on the head's id, but let me know if doing it this way is normal.

src/main/webapp/vendor/startbootstrap/index.html
```diff
@@ -54,7 +54,7 @@
 
       <!-- Nav Item - 4 Year Planner -->
       <li class="nav-item">
-        <a class="nav-link" href="index.html">
+        <a class="nav-link" href="planner.html">
```
savsa, line 57: `/planner` for the href should work, right? That's a lot more standard and better looking.

savsa (changes requested):

src/main/webapp/script.js
ramyabuva: the problem that I saw with that is I wouldn't be able to make those functions and variables I want to import anonymous, which could potentially cause the namespace collisions that William mentioned.

src/main/webapp/vendor/startbootstrap/index.html
ramyabuva: that takes you to the servlet page, not actual html page

src/main/webapp/script.js
wli: Great points from both of you. Some thoughts:

1. Instead of document head id (i've also never seen this before), we could:
  a. do this only when the calendar element (`#calendar`) exists [I would lean towards this if we're going with the 1-javascript-file approach]
  b. inspect the url of the page
2. We chatted a bit about the new file option during our sync today. We can reduce namespace collisions by "owning" a top-level namespace (`CollegePlanner.*`) that is very unlikely to collide with anything. So you could create a `CollegePlanner.CourseSelector` component for example. This is a bit more sustinable since we can just create more files as different use cases come up.

src/main/webapp/vendor/startbootstrap/index.html
savsa: Oh right. It seems off to me but it's fine for now.

src/main/java/com/google/collegeplanner/servlets/PlannerServlet.java
```diff
@@ -0,0 +1,40 @@
...
+import javax.servlet.http.HttpServletResponse;
+
+/** Servlet that returns the mutli-semester plan for the given courses.*/
+@WebServlet("/planner")
+public class PlannerServlet extends HttpServlet {
+  /**
+   * Organizes courses into the given number of semesters
+   */
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
```
savsa: I think this is better suited with `doPost`. GET requests  typically always serve the same the same content every time it's called (like website pages) but POST requests are more complicated and do things that you don't want people to be able to do by just typing in the url in their browser (like logging in, submitting forms, etc).  This servlet will eventually do more complicated things so I think POST is better here.

savsa (approved):

src/main/java/com/google/collegeplanner/servlets/PlannerServlet.java
```diff
@@ -0,0 +1,40 @@
...
+import com.google.gson.Gson;
+import java.io.IOException;
+import java.util.ArrayList;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/** Servlet that returns the mutli-semester plan for the given courses.*/
+@WebServlet("/planner")
```
wli, line 31: @savsa brought up a great point. For any servlet that is part of our API, can we put these paths at `/api/planner` and `/api/courses`, etc?

Then any "html" pages can be `/planner`, etc.

src/main/java/com/google/collegeplanner/servlets/PlannerServlet.java
wli: +1, great callout.

src/main/java/com/google/collegeplanner/servlets/PlannerServlet.java
```diff
@@ -0,0 +1,40 @@
...
+public class PlannerServlet extends HttpServlet {
+  /**
+   * Organizes courses into the given number of semesters
+   */
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    // TODO: Implement algorithm to split into the number of semesters
+    String strClasses = request.getParameter("selectedClasses");
```
wli: can strClasses be null? If so, this can NullPointerException unless you check for it..

src/main/webapp/script.js
```diff
@@ -100,21 +100,84 @@ window.addEventListener('load', () => {
...
+  } else {
+    document.getElementById('submit-plan').addEventListener('submit', () => {
+      /**
+       * Gets results from /planner servlet to results
+       */
+      async function getPlan() {
+        const courseContainer = document.getElementById('order-area');
+        createSpinner(courseContainer);
+        const response = await fetch(
+            `/planner?selectedClasses=${selected.toString()}&semesters=${
```
wli: instead of `toString`, which can change from time to time, i think you want `.join(',')`

wli: To be safe, you should use `encodeURIComponent` for any url params. Otherwise, stray characters like `?` or `&` can mess up the url.

wli: I forgot - if this becomes a doPost, then you'll want to specify these arguments in the request body anyway, so this might be a moot point.

src/main/webapp/script.js
```diff
@@ -100,21 +100,84 @@ window.addEventListener('load', () => {
...
+                document.getElementById('semesters').value}`);
+        courseList = await response.json();
```
wli: where is `courseList` declared? should this be `const courseList`?

src/main/webapp/script.js
```diff
@@ -100,21 +100,84 @@ window.addEventListener('load', () => {
...
+        createTable(courseList, courseContainer);
+      }
+
+      /**
+       * Creates spinner to signify loading
+       * @param {Object} courseContainer container for course list
+       */
+      function createSpinner(courseContainer) {
```
wli: maybe `attachNewSpinner`? The name `createSpinner` does not imply modification to any arguments.

src/main/webapp/script.js
```diff
@@ -100,21 +100,84 @@ window.addEventListener('load', () => {
...
+       * @param {Object} tableData 2D array with separation of courses
+       * @param {Object} courseContainer container for course list
+       */
+      function createTable(tableData, courseContainer) {
+        courseContainer.innerText = '';
+        const table = document.createElement('table');
+        const tableBody = document.createElement('tbody');
+        table.setAttribute('class', 'table table-hover mb-0;');
+        let i = 1;
+        tableData.forEach(function(rowData) {
```
wli: 1. please use arrow functions for consistency
2. You can have an automatic counter

```
tableData.forEach((rowData, i) => {
  // note that i starts at 0, but you can just add 1 to it when you need.
});

File: src/main/webapp/script.js [hunk 14]
```diff
@@ -100,21 +100,84 @@ window.addEventListener('load', () => {
...
+          const row = document.createElement('tr');
+          const cell = document.createElement('td');
+          const semesterLabel = document.createElement('b');
+          semesterLabel.innerText = `Semester ${i}:`;
+          cell.appendChild(semesterLabel);
+          row.appendChild(cell);
+
+          rowData.forEach(function(cellData) {
```
wli: same here, use arrow functions.

src/main/webapp/vendor/startbootstrap/planner.html
```diff
@@ -0,0 +1,250 @@
+<!DOCTYPE html>
```
wli, line 1: i think @savsa was looking into sharing the header/base template for a page. Any chance we could figure that out soon? If not, let's file an issue to handle it soon so we don't have to maintain two versions of this.

wli (changes requested):

src/main/webapp/vendor/startbootstrap/planner.html
wli, line 1: i think this file can be deleted now right, if we're going to render it straight from jsp?

src/main/webapp/vendor/startbootstrap/planner-content.jsp
```diff
@@ -0,0 +1,94 @@
+<%@ taglib uri="http://java.sun.com/jsp/jstl/core" prefix ="c" %> <%@ taglib
```
wli, line 1: can the second `taglib` statement go on its own line?

src/main/webapp/vendor/startbootstrap/index.jsp
```diff
@@ -0,0 +1,3 @@
+<jsp:include page="/vendor/startbootstrap/template.jsp">
+  <jsp:param name="content" value="index-content"/>
+</jsp:include>
```
wli, line 3: nit: newline

src/main/webapp/vendor/startbootstrap/index-content.jsp
```diff
@@ -0,0 +1,80 @@
+<%@ taglib uri="http://java.sun.com/jsp/jstl/core" prefix ="c" %> <%@ taglib
```
wli, line 1: can the second `taglib` statement go on its own line?

src/main/webapp/vendor/startbootstrap/index.html
```diff
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html lang="en">
 
-<head>
+<head id="scheduler">
```
wli, line 4: is this necessary?

src/main/webapp/vendor/startbootstrap/template.jsp
```diff
@@ -0,0 +1,197 @@
+<%@ page language="java" contentType="text/html; charset=UTF-8"
+	pageEncoding="UTF-8"%>
+<%@ taglib uri="http://java.sun.com/jsp/jstl/core" prefix="c"%>
+<!DOCTYPE html>
+<html lang="en">
+
+<head id="scheduler">
```
wli, line 7: can `id="scheduler"` be removed?

wli, line 7: indentation is wrong in this file. please fix.

src/main/webapp/vendor/startbootstrap/template.jsp
```diff
@@ -0,0 +1,197 @@
...
+                </a>
+              </div>
+            </li>
+
+          </ul>
+
+        </nav>
+        <!-- End of Topbar -->
+
+
```
wli, line 128: extra blank line

src/main/webapp/vendor/startbootstrap/template.jsp
```diff
@@ -0,0 +1,197 @@
...
+      <footer class="sticky-footer bg-white">
+        <div class="container my-auto">
+          <div class="copyright text-center my-auto">
+            <span>Copyright &copy; Capstone-Filler</span>
+          </div>
+        </div>
+      </footer>
+      <!-- End of Footer -->
+
+
```
wli, line 143: extra blank line.

makefile
```diff
@@ -8,7 +8,8 @@ node_modules:
 	npm install prettier clang-format css-validator html-validate eslint eslint-config-google
 
 devprep: 
-	cp src/main/webapp/vendor/startbootstrap/index.html src/main/webapp/index.html
+	cp src/main/webapp/vendor/startbootstrap/index.jsp src/main/webapp/index.jsp
```
wli, line 11: do we still need to do this? especially for planner.jsp, I don't think it's necessary since we'll be dynamically serving them.

src/main/java/com/google/collegeplanner/servlets/PlannerServlet.java
```diff
@@ -0,0 +1,70 @@
...
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import org.json.simple.JSONArray;
+import org.json.simple.JSONObject;
+import org.json.simple.parser.JSONParser;
+import org.json.simple.parser.ParseException;
+
+/** Servlet that returns the mutli-semester plan for the given courses.*/
+@WebServlet("/planner")
```
wli, line 31: /api/planner

src/main/webapp/script.js
```diff
@@ -100,21 +100,89 @@ window.addEventListener('load', () => {
...
+          row.appendChild(cell);
+
+          rowData.forEach((cellData) => {
+            const cell = document.createElement('td');
+            cell.appendChild(document.createTextNode(cellData));
+            row.appendChild(cell);
+          });
+
+          tableBody.appendChild(row);
+          i += 1;
```
wli, line 178: this line is unnecessary

wli (changes requested):

PR #19, opened 2020-07-08
Add UMD API

savsa:


src/main/java/com/google/collegeplanner/servlets/APIUtil.java
```diff
@@ -0,0 +1,124 @@
...
+      jsonObject = (JSONObject) parser.parse(json);
+    } catch (ParseException e) {
+      return null;
+    }
+    return jsonObject;
+  }
+
+  /**
+   * Makes a GET request and returns the reponse json.
+
```
ramyabuva: nit: remove additional line or put *

ramyabuva (approved):

src/main/java/com/google/collegeplanner/servlets/APIUtil.java
```diff
@@ -0,0 +1,124 @@
...
+import org.apache.http.impl.client.HttpClients;
+import org.apache.http.ssl.SSLContextBuilder;
+import org.apache.http.util.EntityUtils;
+import org.json.simple.JSONArray;
+import org.json.simple.JSONObject;
+import org.json.simple.parser.JSONParser;
+import org.json.simple.parser.ParseException;
+
+/** Provides an interface for making outside API calls. */
+public class APIUtil {
```
wli: This should be ApiUtil: https://google.github.io/styleguide/javaguide.html#s5.3-camel-case

Or maybe even UmdApiUtil?

src/main/java/com/google/collegeplanner/servlets/APIUtil.java
```diff
@@ -0,0 +1,124 @@
...
+  /**
+   * Returns a json object given a URIBuilder.
+   *
+   * @param builder The endpoint that will be requested
+   */
+  public JSONArray requestAPIAndGetJsonArray(URIBuilder builder) {
```
wli: since we're not modifying the URI inside this function, this can just take a full `URI` instead of `URIBuilder`

wli: What about `getJsonArray` as the function name? The full call becomes `ApiUtil.getJsonArray()`, which seems clear enough that it is about getting a json array from an API.

src/main/java/com/google/collegeplanner/servlets/APIUtil.java
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    HttpGet apiRequest;
+    try {
+      apiRequest = new HttpGet(builder.build());
+    } catch (URISyntaxException e) {
+      return null;
+    }
+
+    String json = requestAPI(apiRequest);
```
wli: why not put lines 74-83 inside of `requestApi` too?

src/main/java/com/google/collegeplanner/servlets/APIUtil.java
```diff
@@ -0,0 +1,124 @@
...
+    return jsonObject;
+  }
+
+  /**
+   * Makes a GET request and returns the reponse json.
+
+   * @param apiRequest The HttpGet object that will be executed
+   */
+  private String requestAPI(HttpGet apiRequest) {
+    // Ignore any SSL certificate validations.
```
wli: we shouldn't ignore this forever - let's make a note why we're doing this, and when it is safe to undo this.

src/main/java/com/google/collegeplanner/servlets/APIUtil.java
```diff
@@ -0,0 +1,124 @@
...
+    try (CloseableHttpClient httpClient =
+             HttpClients.custom()
+                 .setSSLContext(new SSLContextBuilder()
+                                    .loadTrustMaterial(null, TrustAllStrategy.INSTANCE)
+                                    .build())
+                 .setSSLHostnameVerifier(NoopHostnameVerifier.INSTANCE)
+                 .build();
+         CloseableHttpResponse apiResponse = httpClient.execute(apiRequest)) {
+      HttpEntity entity = apiResponse.getEntity();
+      json = EntityUtils.toString(entity);
```
wli: what about this:

```
try {
  // ... 
  if (entity != null && apiResponse... == SC_OK) {
    return EntityUtils.toString(entity);
  }
} catch {
}

return "";
```

Then you can get rid of the `String json;` declaration on line 103, and all your error/failure cases condense down into a single return line.

src/main/java/com/google/collegeplanner/servlets/CourseListServlet.java
```diff
@@ -31,21 +35,37 @@
...
 
     JSONObject schoolCourseInfo = new JSONObject();
-    schoolCourseInfo.put("courses_detailed", courses);
+    schoolCourseInfo.put("courses_detailed", jsonArray);
 
     response.setContentType("applications/json;");
     response.getWriter().println(schoolCourseInfo);
   }
+
+  private void respondWithError(String message, int errorType, HttpServletResponse response)
```
wli, line 75: No need to do it in this PR specifically, but I think we should create a `CollegePlannerServlet` which contains the `respondWithError` function, and then this class could be `class CourseListServlet extends CollegePlannerServlet`. That way, we can be consistent with our errors, and share other potential utils/functions/etc.

Please feel free to file an issue and follow up on this later.

src/test/java/com/google/collegeplanner/TestCourseList.java
```diff
@@ -36,9 +36,44 @@ public void servletResponseHasCourses() throws Exception {
...
-    Assert.assertEquals(coursesDetailed.size(), 18);
+    Assert.assertEquals(coursesDetailed.size(), 30);
     // Checks whether the first one is CMSC101
     String firstCourse = coursesDetailed.get(0).toString();
-    Assert.assertEquals(firstCourse, "{\"name\":\"CMSC101\"}");
+    String expectedJson = "{"
+        + "\"course_id\":\"AASP100\","
+        + "\"core\":[\"SH\",\"D\"],"
+        + "\"relationships\":{"
+        + "\"coreqs\":null,"
```
wli: can you indent here so it's slightly easier for humans to see the structure?

```
+ "\"relationships\":{"
+ "  \"coreqs\": null,"
```

wli (changes requested):

src/test/java/com/google/collegeplanner/TestCourseList.java
savsa: I did indent everything properly but the auto formatter changes it to this.

wli: wait, the formatter removed whitespace inside of the string? That doesn't seem right. I could see that it would remove whitespace outside the string.

ramyabuva: If he changes the string itself, will it still match with the JSON string retrieved from the servlet?

savsa: Oh sorry, I saw that wrong. I added tabs to format the code, not inside the strings. But anyways, adding spaces to the string would cause the equals-assertion to fail, right?

wli: Is there an easy "JsonEquals" assertion we can use that is whitespace/formatting-agnostic?

wli: If so, we could do:

assertJsonEquals(coursesDetailed.get(0), Json.parse(expectedJson));

savsa: Found a third-party library to do this: https://github.com/skyscreamer/JSONassert

src/main/java/com/google/collegeplanner/servlets/CourseListServlet.java
savsa: Filed an issue.

src/main/java/com/google/collegeplanner/servlets/ApiUtil.java
```diff
@@ -0,0 +1,109 @@
...
+import org.json.simple.JSONObject;
+import org.json.simple.parser.JSONParser;
+import org.json.simple.parser.ParseException;
+
+/** Provides an interface for making outside API calls. */
+public class ApiUtil {
+  /**
+   * Returns a json object given a URI.
+   *
+   * @param builder The endpoint that will be requested
```
wli: comment is out of date

src/main/java/com/google/collegeplanner/servlets/CourseListServlet.java
```diff
@@ -31,21 +35,38 @@
...
+    } catch (URISyntaxException e) {
+      respondWithError(
+          "Internal server error.", HttpServletResponse.SC_INTERNAL_SERVER_ERROR, response);
+      return;
+    }
+
+    JSONArray jsonArray = new ApiUtil().getJsonArray(uri);
+    if (jsonArray == null) {
+      respondWithError(
+          "Internal server error.", HttpServletResponse.SC_INTERNAL_SERVER_ERROR, response);
```
wli, line 64: i think you need to immediately return if this fails.

src/test/java/com/google/collegeplanner/TestCourseList.java
```diff
@@ -36,9 +38,43 @@ public void servletResponseHasCourses() throws Exception {
...
+        + "  \"AASP100-0101\","
+        + "  \"AASP100-0201\","
+        + "  \"AASP100-0301\","
+        + "  \"AASP100-0401\","
+        + "  \"AASP100-0501\","
+        + "  \"AASP100-0601\","
+        + "  \"AASP100-0701\""
+        + "]"
+        + "}";
+    JSONAssert.assertEquals(coursesDetailed.get(0).toString(), expectedJson, JSONCompareMode.STRICT);
```
wli: missing test cases for the error paths.

wli (changes requested):

src/main/java/com/google/collegeplanner/servlets/ApiUtil.java
```diff
@@ -0,0 +1,109 @@
...
+                 .setSSLContext(new SSLContextBuilder()
+                                    .loadTrustMaterial(null, TrustAllStrategy.INSTANCE)
+                                    .build())
+                 .setSSLHostnameVerifier(NoopHostnameVerifier.INSTANCE)
+                 .build();
+         CloseableHttpResponse apiResponse = httpClient.execute(apiRequest)) {
+      HttpEntity entity = apiResponse.getEntity();
+      String json = EntityUtils.toString(entity);
+      if (entity != null
+          || apiResponse.getStatusLine().getStatusCode() == HttpServletResponse.SC_OK) {
```
triplehair: Should this be && ?

src/main/java/com/google/collegeplanner/servlets/ApiUtil.java
```diff
@@ -0,0 +1,109 @@
...
+    } catch (ParseException e) {
+      return null;
+    }
+    return jsonArray;
+  }
+
+  /**
+   * Returns a json object given a URIBuilder.
+   *`
+   * @param builder The endpoint that will be requested
```
triplehair: Also update this and and the one below

triplehair (changes requested):

wli (approved):
Great work!

triplehair (approved):

PR #30, opened 2020-07-15
Implement planner algorithm

ramyabuva:
https://screenshot.googleplex.com/fRfbOi433S1.png

### DIFF
--- a/src/main/java/com/google/collegeplanner/servlets/PlannerServlet.java
+++ b/src/main/java/com/google/collegeplanner/servlets/PlannerServlet.java
@@ -36,7 +36,8 @@ import org.json.simple.parser.ParseException;
 @WebServlet("/api/planner")
 public class PlannerServlet extends HttpServlet {
   private HashMap<String, Integer> indegree; // Number of prerequisites for a course
-  private HashMap<String, HashSet<String>> nextCourses; // Courses that come after the given course in the graph
+  private HashMap<String, HashSet<String>>
+      nextCourses; // Courses that come after the given course in the graph
   private HashMap<String, HashSet<String>> corequisites;
   private HashMap<String, Integer> depths; // depth of a course in the graph
   private HashMap<String, Integer> credits; // Number of credits for each course
@@ -109,6 +110,8 @@ public class PlannerServlet extends HttpServlet {
 
   /**
    * Given a course, returns whether all of its corequisites have a indegree of 0
+   *
+   * @param course The course ID for the course
    */
   private boolean isValidToAdd(String course) {
     for (String coreq : corequisites.get(course)) {
@@ -133,6 +136,9 @@ public class PlannerServlet extends HttpServlet {
   /**
    * Creates the representation for the graph and reads/stores necessary information from
    * selectedClasses
+   *
+   * @param selectedClasses JSONArray that contains details for all of the courses selected by the
+   *     user
    */
   private void constructGraphs(JSONArray selectedClasses)
       throws IOException, NumberFormatException {
@@ -174,8 +180,8 @@ public class PlannerServlet extends HttpServlet {
     fillDepths();
 
     // sort courseList from greatest depth to lowest depth
-    Collections.sort(
-        courseList, (String course1, String course2) -> depths.get(course2).compareTo(depths.get(course1)));
+    Collections.sort(courseList,
+        (String course1, String course2) -> depths.get(course2).compareTo(depths.get(course1)));
   }
 
   /**
@@ -192,6 +198,8 @@ public class PlannerServlet extends HttpServlet {
 
   /**
    * Dynamic Programming Algorithm to find the depth of a course
+   *
+   * @param course The course ID for the course
    */
   private int getDepth(String course) {
     if (depths.containsKey(course)) {
@@ -213,6 +221,8 @@ public class PlannerServlet extends HttpServlet {
   /**
    * Takes English representation of course prerequisites/corequisites and returns a HashSet with
    * all the contained courses
+   *
+   * @param engCourses The natural language string to get courses from
    */
   private HashSet<String> getCoursesFromString(String engCourses) throws IOException {
     HashSet<String> prereqs = new HashSet<>();
--- a/src/main/java/com/google/collegeplanner/servlets/PlannerServlet.java
+++ b/src/main/java/com/google/collegeplanner/servlets/PlannerServlet.java
@@ -38,7 +38,7 @@ public class PlannerServlet extends HttpServlet {
   static HashMap<String, Integer> indegree;
   static HashMap<String, HashSet<String>> corequisites;
   static HashMap<String, HashSet<String>> postrequisites;
-  static HashMap<String, Integer> depths;
+  static HashMap<String, Integer> depths; // depth of a course (number in a prerequisite "chain" begining at that node)
   static HashMap<String, Integer> credits;
   static int totalCredits;
   static ArrayList<String> courseList;
@@ -48,12 +48,11 @@ public class PlannerServlet extends HttpServlet {
    */
   @Override
   public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
-    // TODO(ramyabuva): Implement algorithm to split into the number of semesters
     JSONObject body;
     int numSemesters;
     try {
       body = getBody(request);
-      numSemesters = Integer.parseInt((String)body.get("semesters"));
+      numSemesters = Integer.parseInt((String) body.get("semesters"));
     } catch (ParseException | NullPointerException | NumberFormatException e) {
       respondWithError(
           "Invalid body for POST request.", HttpServletResponse.SC_BAD_REQUEST, response);
@@ -67,27 +66,47 @@ public class PlannerServlet extends HttpServlet {
     for (int i = numSemesters; i > 0; i--) {
       int avgCredits = totalCredits / i;
       ArrayList<String> semester = new ArrayList<>();
-      Iterator<String> itr = courseList.iterator(); 
-      while (itr.hasNext()) { 
-        String course = itr.next(); 
-        if (indegree.get(course) == 0) { 
-          itr.remove();
-          avgCredits -= credits.get(course);
-          totalCredits -= credits.get(course);
-          semester.add(course); 
+      for (int j = 0; j < courseList.size(); j++) {
+        String course = courseList.get(j);
+        if (indegree.get(course) == 0) {
+          // Check if all corequisites also have an indegree of 0
+          boolean isValid = true;
+          for (String coreq : corequisites.get(course)) {
+            if (indegree.get(coreq) != 0) {
+              isValid = false;
+              break;
+            }
+          }
+          if (isValid) {
+            // If valid, add all to semester
+            courseList.remove(j);
+            j--;
+            avgCredits -= credits.get(course);
+            totalCredits -= credits.get(course);
+            semester.add(course);
+            for (String coreq : corequisites.get(course)) {
+              courseList.remove(coreq);
+              avgCredits -= credits.get(coreq);
+              totalCredits -= credits.get(coreq);
+              semester.add(coreq);
+            }
+          }
         } 
-        if(depths.get(course) != i && avgCredits <= 0) {
+        // Breaks out if there are no more required classes and we exceeded our credit limit
+        if (depths.get(course) != i && avgCredits <= 0) {
           break;
         }
       }
+      // Update indegrees for all postrequisites
       for (String course : semester) {
         for (String postreq : postrequisites.get(course)) {
           indegree.replace(postreq, indegree.get(postreq) - 1);
         }
-      } 
+      }
       semesters.add(semester);
     }
-    if(courseList.size() != 0){
+    // If courseList is not empty, scheduling was not possible
+    if (!courseList.isEmpty()) {
       respondWithError(
           "Not enough semesters for schedule.", HttpServletResponse.SC_BAD_REQUEST, response);
     }
@@ -105,7 +124,7 @@ public class PlannerServlet extends HttpServlet {
 
   private static void constructGraphs(JSONArray selectedClasses)
       throws IOException, NumberFormatException {
-    // Initialize maps and courseList
+    // Initialize graph representation and courseList
     indegree = new HashMap<String, Integer>();
     corequisites = new HashMap<String, HashSet<String>>();
     depths = new HashMap<String, Integer>();
@@ -139,8 +158,8 @@ public class PlannerServlet extends HttpServlet {
     getAllDepths();
 
     // sort courseList from greatest depth to lowest depth
-    Collections.sort(courseList, (String m1, String m2) -> 
-            depths.get(m2).compareTo(depths.get(m1)));
+    Collections.sort(
+        courseList, (String m1, String m2) -> depths.get(m2).compareTo(depths.get(m1)));
   }
 
   private static void getAllDepths() {
@@ -169,8 +188,7 @@ public class PlannerServlet extends HttpServlet {
     return myHeight;
   }
 
-  private static HashSet<String> getCoursesFromString(String engCourses)
-      throws IOException {
+  private static HashSet<String> getCoursesFromString(String engCourses) throws IOException {
     HashSet<String> prereqs = new HashSet<>();
     try {
       String[] words = engCourses.split("[\\p{Punct}\\s]+");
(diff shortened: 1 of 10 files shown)


src/main/java/com/google/collegeplanner/servlets/PlannerServlet.java
```diff
@@ -29,35 +35,199 @@
...
         request.getReader().lines().collect(Collectors.joining(System.lineSeparator()));
     JSONParser parser = new JSONParser();
     return (JSONObject) parser.parse(strBody);
   }
 
+  /**
+   * Creates the representation for the graph and reads/stores necessary information from
+   * selectedClasses
+   */
+  private void constructGraphs(JSONArray selectedClasses)
```
savsa, line 153: All the java functions you're adding should have a brief description of each parameter

savsa (approved):

src/main/java/com/google/collegeplanner/servlets/PlannerServlet.java
```diff
@@ -29,35 +35,209 @@
...
       return;
     }
-    String strClasses = (String) body.get("selectedClasses");
-    // Returns the courses in a single semester
-    String[][] courses = {strClasses.split(",")};
+    JSONArray selectedClasses = (JSONArray) body.get("selectedClasses");
+
+    constructGraphs(selectedClasses);
+
+    ArrayList<ArrayList<String>> semesters = new ArrayList<>();
```
wli, line 80: there's a lot going on in this section. I would pull the following code out into its own helper function so its easier to follow.

src/main/java/com/google/collegeplanner/servlets/PlannerServlet.java
```diff
@@ -29,35 +35,209 @@
...
+      numSemesters = Integer.parseInt((String) body.get("semesters"));
+    } catch (ParseException | NullPointerException | NumberFormatException e) {
       respondWithError(
           "Invalid body for POST request.", HttpServletResponse.SC_BAD_REQUEST, response);
       return;
     }
-    String strClasses = (String) body.get("selectedClasses");
-    // Returns the courses in a single semester
-    String[][] courses = {strClasses.split(",")};
+    JSONArray selectedClasses = (JSONArray) body.get("selectedClasses");
```
wli: it doesn't look like you validated this input. what happens when you get a garbage input? will the program crash (it should return a 400)? are there potential security issues?

src/main/java/com/google/collegeplanner/servlets/PlannerServlet.java
```diff
@@ -29,35 +35,209 @@
...
+    indegree = new HashMap<String, Integer>();
+    corequisites = new HashMap<String, HashSet<String>>();
+    depths = new HashMap<String, Integer>();
+    credits = new HashMap<String, Integer>();
+    nextCourses = new HashMap<String, HashSet<String>>();
+    courseList = new ArrayList<>();
+
+    // creates CourseList and initialize nextCourses HashSets
+    for (Object course : selectedClasses) {
+      String key = (String) ((JSONObject) course).get("course_id");
```
wli, line 166: it seems like you're doing `(JSONObject) course` a lot. You can pull that out into a temporary variable so the rest of the code is simpler to read. In the near term, we should actually use Course.java that @naaoli is writing instead of passing a json object.

src/main/java/com/google/collegeplanner/servlets/PlannerServlet.java
```diff
@@ -29,35 +35,209 @@
...
+      courseList.add(key);
+      nextCourses.put(key, new HashSet<>());
+    }
+
+    // populate class variables
+    for (Object course : selectedClasses) {
+      String key = (String) ((JSONObject) course).get("course_id");
+      int creditVal = Integer.parseInt((String) ((JSONObject) course).get("credits"));
```
wli: It seems like you're getting the full data object from the frontend. Note that this is not really best practice. Consider this example:

1. User selects courses.
2. Database updates with new prerequisite data.
3. User generates a schedule.

The user will wind up using out-of-date data. Instead, we should only be passing the course_id as an input, and we should do a database lookup.

Of course, we'll need to defer this until we get datastore working though, but it would be good to file an issue to handle this.

src/main/java/com/google/collegeplanner/servlets/PlannerServlet.java
```diff
@@ -29,35 +35,209 @@
...
+    return courseDepth;
+  }
+
+  /**
+   * Takes English representation of course prerequisites/corequisites and returns a HashSet with
+   * all the contained courses
+   *
+   * @param engCourses The natural language string to get courses from
+   */
+  private HashSet<String> getCoursesFromString(String engCourses) throws IOException {
```
wli: we try not to unnecessarily abbreviate unless it's super common (like http).

This should be `engCourses` -> `englishCourses`. I initially read this as `engineeringCourses`.

Or maybe it would be even clearer as `courseRelationshipsDescription` or something? Since `*Courses` implies the Course.java object that we're about to add.

src/main/java/com/google/collegeplanner/servlets/PlannerServlet.java
```diff
@@ -29,35 +35,209 @@
...
+        for (String nextCourse : nextCourses.get(course)) {
+          indegree.replace(nextCourse, indegree.get(nextCourse) - 1);
+        }
+      }
+      semesters.add(semester);
+    }
+    // If courseList is not empty, scheduling was not possible
+    if (!courseList.isEmpty()) {
+      respondWithError(
+          "Not enough semesters for schedule.", HttpServletResponse.SC_BAD_REQUEST, response);
```
wli: if you respond with an error, do you need to return? Is it ok if you print out `semesters` below in addition?

src/main/java/com/google/collegeplanner/servlets/PlannerServlet.java
```diff
@@ -29,35 +35,209 @@
...
+      // Update indegrees for all next courses
+      for (String course : semester) {
+        for (String nextCourse : nextCourses.get(course)) {
+          indegree.replace(nextCourse, indegree.get(nextCourse) - 1);
+        }
+      }
+      semesters.add(semester);
+    }
+    // If courseList is not empty, scheduling was not possible
+    if (!courseList.isEmpty()) {
```
wli: minor point, but i think it's ok to return status code 200 with 0 results in this case.

src/test/java/com/google/collegeplanner/PlannerTest.java
```diff
@@ -37,9 +37,50 @@
 @RunWith(JUnit4.class)
 public final class PlannerTest {
```
wli, line 46: missing failure test cases:

 - what happens when no schedule can be made due to conflicts / not enough credits or semesters
 - what happens when no schedule can be made due to an infinite prereq loop
 - what happens when you have an empty input? or invalid input?
 - ...and any other unique failure mode you can think of

wli (changes requested):

src/main/java/com/google/collegeplanner/servlets/PlannerServlet.java
```diff
@@ -29,35 +35,221 @@
...
     JSONParser parser = new JSONParser();
     return (JSONObject) parser.parse(strBody);
   }
 
+  /**
+   * Creates the representation for the graph and reads/stores necessary information from
+   * selectedClasses
+   *
+   * @param selectedClasses JSONArray that contains details for all of the courses selected by the
+   *     user
```
triplehair, line 151: remove extra spaces

src/main/java/com/google/collegeplanner/servlets/PlannerServlet.java
```diff
@@ -29,35 +35,221 @@
...
+    // Initialize graph representation and courseList
+    indegree = new HashMap<String, Integer>();
+    corequisites = new HashMap<String, HashSet<String>>();
+    depths = new HashMap<String, Integer>();
+    credits = new HashMap<String, Integer>();
+    nextCourses = new HashMap<String, HashSet<String>>();
+    courseList = new ArrayList<>();
+
+    // creates CourseList and initialize nextCourses HashSets
+    for (Object course : selectedClasses) {
```
triplehair, line 165: Can there be duplicated courses in selectedClasses?

src/main/java/com/google/collegeplanner/servlets/PlannerServlet.java
```diff
@@ -29,35 +35,221 @@
...
+   */
+  private int getDepth(String course) {
+    if (depths.containsKey(course)) {
+      return depths.get(course);
+    }
+    if (nextCourses.get(course).size() == 0) {
+      depths.put(course, 1);
+      return 1;
+    }
+    ArrayList<Integer> childrenDepths = new ArrayList<>();
```
triplehair: Instead of ArrayList, you can just have `int maxDepth` and in the for loop, update maxDepth if you see a larger value.

src/main/java/com/google/collegeplanner/servlets/PlannerServlet.java
```diff
@@ -29,35 +35,221 @@
...
+      int avgCredits = totalCredits / i;
+      ArrayList<String> semester = new ArrayList<>();
+      for (int j = 0; j < courseList.size(); j++) {
+        String course = courseList.get(j);
+        if (indegree.get(course) == 0) {
+          // Check if all corequisites also have an indegree of 0
+          if (isValidToAdd(course)) {
+            // If valid, add all to semester
+            courseList.remove(j);
+            j--;
```
triplehair, line 94: It is error prone to update j in the body of the for loop as well as the condition of for loop. Since you are removing elements from courseList, courseList.size() is changing, consider using while(courseList.size() >= 0) as the loop.

src/main/java/com/google/collegeplanner/servlets/PlannerServlet.java
```diff
@@ -29,35 +35,221 @@
...
+            for (String coreq : corequisites.get(course)) {
+              courseList.remove(coreq);
+              avgCredits -= credits.get(coreq);
+              totalCredits -= credits.get(coreq);
+              semester.add(coreq);
+            }
+          }
+        }
+        // Breaks out if there are no more required classes and we exceeded our credit limit
+        if (depths.get(course) != i && avgCredits <= 0) {
```
triplehair: This is smart! With `depths.get(course) != i`, we are sure we can come up with a plan. Maybe add some comments to explain this?

src/main/java/com/google/collegeplanner/servlets/PlannerServlet.java
```diff
@@ -29,35 +35,221 @@
...
+    for (int i = numSemesters; i > 0; i--) {
+      int avgCredits = totalCredits / i;
+      ArrayList<String> semester = new ArrayList<>();
+      for (int j = 0; j < courseList.size(); j++) {
+        String course = courseList.get(j);
+        if (indegree.get(course) == 0) {
+          // Check if all corequisites also have an indegree of 0
+          if (isValidToAdd(course)) {
+            // If valid, add all to semester
+            courseList.remove(j);
```
triplehair, line 93: This takes O(n) to remove an element from an ArrayList, and given that j starts from 0, so from time complexity point of view, it would be the worst case.
If we define courseList as a LinkedList, remove takes O(1).

src/main/java/com/google/collegeplanner/servlets/PlannerServlet.java
```diff
@@ -29,35 +35,221 @@
 /** Servlet that returns the mutli-semester plan for the given courses.*/
 @WebServlet("/api/planner")
 public class PlannerServlet extends HttpServlet {
+  private HashMap<String, Integer> indegree; // Number of prerequisites for a course
+  private HashMap<String, HashSet<String>>
+      nextCourses; // Courses that come after the given course in the graph
+  private HashMap<String, HashSet<String>> corequisites;
+  private HashMap<String, Integer> depths; // depth of a course in the graph
```
triplehair: The name depths is a little confusing. Usually tree's root node has depth 0, while tree's leaf has height 0. https://stackoverflow.com/questions/2603692/what-is-the-difference-between-tree-depth-and-height.
From the code below, is it more like the height of a tree?

src/main/java/com/google/collegeplanner/servlets/PlannerServlet.java
```diff
@@ -29,35 +35,221 @@
...
+    response.getWriter().println(new Gson().toJson(semesters));
   }
 
-  private static JSONObject getBody(HttpServletRequest request)
+  /**
+   * Creates and returns a valid plan for the given number of semesters
+   *
+   * @param numSemesters The number of semesters remaining
+   */
+  private ArrayList<ArrayList<String>> getPlan(int numSemesters) {
```
triplehair, line 79: It won't be possible to have a plan if max depths > numSemesters, right? We can early return an error in this case.

triplehair (changes requested):

src/main/java/com/google/collegeplanner/servlets/PlannerServlet.java
```diff
@@ -29,35 +35,221 @@
...
-    String strClasses = (String) body.get("selectedClasses");
-    // Returns the courses in a single semester
-    String[][] courses = {strClasses.split(",")};
+
+    constructGraphs(selectedClasses);
+
+    ArrayList<ArrayList<String>> semesters = getPlan(numSemesters);
+    // If courseList is not empty, scheduling was not possible
+    if (!courseList.isEmpty()) {
+      return;
```
wli: looks like you're returning an empty body, instead of an empty json array. Let's make sure you have at least something in the return output.

In addition, I would wrap the output in an Object, so it's more future-proof:

```
{
  semesters: ...
  some_other_future_field: ...
}
```

wli (changes requested):

src/main/java/com/google/collegeplanner/servlets/PlannerServlet.java
ramyabuva: I was thinking of the end of the graph as the root node, but I agree that it makes more sense to use height instead

src/main/java/com/google/collegeplanner/servlets/PlannerServlet.java
ramyabuva: I added the extra spaces to stay consistent with the comments in Savion's ApiUtil class; should I remove them from both classes?

src/main/java/com/google/collegeplanner/servlets/PlannerServlet.java
ramyabuva: I considered this, but I realized I would still be maintaining a variable to iterate through the list, and you don't want to repeat until the courseList is empty, but some other conditions. Also, since j begins at 0 and is never less than 0 when the condition is evaluated, this for loop should never produce an error.

savsa (approved):

src/main/java/com/google/collegeplanner/servlets/PlannerServlet.java
```diff
@@ -29,35 +36,230 @@
...
-    // TODO(ramyabuva): Implement algorithm to split into the number of semesters
     JSONObject body;
+    int numSemesters;
+    JSONArray selectedClasses;
     try {
       body = getBody(request);
-    } catch (ParseException | NullPointerException e) {
+      numSemesters = Integer.parseInt((String) body.get("semesters"));
+      selectedClasses = (JSONArray) body.get("selectedClasses");
+    } catch (Throwable e) {
```
wli: generally, you should be catching concrete exceptions: https://screenshot.googleplex.com/RZ3ac61r01n

wli (approved):

wli (approved):

triplehair (approved):

src/main/java/com/google/collegeplanner/servlets/PlannerServlet.java
triplehair: Yes I think you should remove in both files. Does lint catch this?

triplehair (approved):

src/main/java/com/google/collegeplanner/servlets/PlannerServlet.java
ramyabuva: Unfortunately not, but I just went back and changed it.

wli:
Restored this branch so that we have history of this code and can add it back it correctly afterwards.

PR #42, opened 2020-07-17
Create BaseServlet class

### LINKED ISSUE #32: Create BaseServlet class
Have an ErrorUtil class to house `respondWithError`

savsa:
The BaseServlet class houses some common logic and functionality between our servlets such as `respondWithError()` and `ApiUtil`.


src/main/java/com/google/collegeplanner/servlets/PlannerServlet.java
```diff
@@ -39,8 +39,7 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
     try {
       body = getBody(request);
     } catch (ParseException | NullPointerException e) {
-      respondWithError(
-          "Invalid body for POST request.", HttpServletResponse.SC_BAD_REQUEST, response);
+      respondWithError(HttpServletResponse.SC_BAD_REQUEST, response);
       return;
     }
```
ramyabuva: When you merge master into this branch, there may be one more respondWithError you'll have to change

src/main/java/com/google/collegeplanner/servlets/BaseServlet.java
```diff
@@ -0,0 +1,75 @@
...
+    this.apiUtil = new ApiUtil();
+  }
+
+  public BaseServlet(ApiUtil apiUtil) {
+    this.apiUtil = apiUtil;
+  }
+
+  /**
+   * Responds with an HTTP error.
+   *
```
ramyabuva: remove gap for consistency with other files

ramyabuva (approved):

src/main/java/com/google/collegeplanner/servlets/PlannerServlet.java
savsa: Noted

src/main/java/com/google/collegeplanner/servlets/BaseServlet.java
```diff
@@ -0,0 +1,74 @@
...
+        break;
+      default:
+        message = "Error.";
+    }
+
+    JSONObject jsonObject = new JSONObject();
+    jsonObject.put("message", message);
+    jsonObject.put("status", "error");
+    response.setStatus(status);
+    response.setContentType("applications/json;");
```
wli: `application/json` is the correct version

src/main/java/com/google/collegeplanner/servlets/CourseListServlet.java
```diff
@@ -47,7 +45,7 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
     // Create the URI and specify the parameters.
     String department = request.getParameter("department");
     if (department == null || department == "") {
-      respondWithError("Invalid query parameters.", HttpServletResponse.SC_BAD_REQUEST, response);
+      respondWithError(HttpServletResponse.SC_BAD_REQUEST, response);
```
wli: we lost the "reasoning" here in the API response. Can we make a separate function that takes in a custom error message?

src/main/webapp/js/calendar.js
```diff
@@ -17,19 +17,19 @@
 // TODO(#34): Add ability to choose which section the user wants, not just
```
wli: was this file meant to be included? It's not mentioned in the PR description.

wli (changes requested):

src/main/webapp/js/calendar.js
savsa: It's the auto formatter. I don't know why it chooses to format files now but not before

src/main/java/com/google/collegeplanner/servlets/CourseListServlet.java
savsa: I agree that a version with a custom message would be useful, but not really here since this is just "bad request"?

wli: It does matter here. This is an API, and it should tell the user (our frontend client) what they can do to fix it. Just "bad request" is not enough to narrow down the problem. For this case, "Invalid or missing department" would be a good error message.

src/main/webapp/js/calendar.js
savsa: I can take it out of this PR but some other random PR in the future will get this same problem when they try to auto format.

wli: In cases like this, the right approach is to update this in a separate PR (since it does need to be fixed).

src/main/java/com/google/collegeplanner/servlets/BaseServlet.java
```diff
@@ -0,0 +1,90 @@
...
+        break;
+      case HttpServletResponse.SC_INTERNAL_SERVER_ERROR:
+        // 500
+        message = "Internal server error.";
+        break;
+      default:
+        message = "Error.";
+    }
+
+    JSONObject jsonObject = new JSONObject();
```
wli: you should do `return respondWithError(message, status, response)` to reduce duplication.

src/main/java/com/google/collegeplanner/servlets/SectionServlet.java
```diff
@@ -45,23 +45,21 @@ public SectionServlet(ApiUtil apiUtil) {
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
     String courseId = request.getParameter("course_id");
     if (courseId == null || courseId == "") {
-      respondWithError("Bad request.", HttpServletResponse.SC_BAD_REQUEST, response);
+      respondWithError(HttpServletResponse.SC_BAD_REQUEST, response);
```
wli: same here - "Invalid or missing course id"

wli (approved):

triplehair (approved):

PR #48, opened 2020-07-20
Fix cloud build deployment of master branch

ramyabuva:
On the cloudbuild, there was an issue building the docker container, so instead i replaced it with a different tagged image and removed the dockerfile.

### DIFF
--- a/cloudbuildMaster.yaml
+++ b/cloudbuildMaster.yaml
@@ -1,5 +1,4 @@
 steps:
- - name: 'gcr.io/cloud-builders/docker'
-   args: ['build', '-t', 'gcr.io/step-team-145/image', '.']
- - name: 'gcr.io/step-team-145/image'
-   args: ['deploy']
+ - name: maven:3-jdk-8
+   entrypoint: mvn
+   args: ['package', 'appengine:deploy']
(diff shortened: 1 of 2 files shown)


wli:
What was the problem + fix? Can you write it in the PR description so it's clear?

Also, why are there two cloud build checks, and only one ran?

image

ramyabuva:
> What was the problem + fix? Can you write it in the PR description so it's clear?
> 
> Also, why are there two cloud build checks, and only one ran?
> 
> image

Just updated the PR description. The 2nd test is a trigger I created on cloudbuild to see if deployment worked, however I deleted it before making this PR, which is why the check won't run.

cloudbuildMaster.yaml
```diff
@@ -1,5 +1,4 @@
 steps:
- - name: 'gcr.io/cloud-builders/docker'
-   args: ['build', '-t', 'gcr.io/step-team-145/image', '.']
- - name: 'gcr.io/step-team-145/image'
-   args: ['deploy']
+ - name: maven:3-jdk-8
+   entrypoint: mvn
+   args: ['package', 'appengine:deploy']
```
wli: nit: missing newline

wli (approved):

triplehair (approved):

savsa (approved):

PR #49, opened 2020-07-20
Add number of credits per semester to PlannerServlet

### LINKED ISSUE #35: Display number of credits for a given semester
Currently we just display the courses in a table

ramyabuva:
Adds the key 'semester_credits' to the returned JSONObject which holds an array of the number of credits in the created semester plan. 

Closes #35

### DIFF
--- a/src/main/webapp/js/calendar.js
+++ b/src/main/webapp/js/calendar.js
@@ -17,19 +17,19 @@
 // TODO(#34): Add ability to choose which section the user wants, not just
 //  automatically choosing the first section
 
-/** 
+/**
  * The id of the next schedule that will be added to the calendar.
  * @type {number}
  */
 let id = 1;
 
-/** 
+/**
  * The calendar object.
  * @type {Calendar}
  */
 let calendar;
 
-/** 
+/**
  * An alias for the Luxon DateTime object.
  * @type {DateTime}
  */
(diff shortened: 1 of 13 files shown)


src/main/webapp/js/calendar.js
```diff
@@ -80,54 +80,54 @@ function initCalendar() {
...
-    if (meetingDays.includes('Th')) {
-      addCourseToCalendar(course, startTime, endTime, enumDays.DATE_THURSDAY);
-    }
-    if (meetingDays.includes('F')) {
-      addCourseToCalendar(course, startTime, endTime, enumDays.DATE_FRIDAY);
-    }
-    if (meetingDays.includes('Sa')) {
-      addCourseToCalendar(course, startTime, endTime, enumDays.DATE_SATURDAY);
-    }
+/**
```
savsa: If possible, remove these js files from the PR. I know that it's annoying because the auto formatter picks up on these, but william said that we should create a separate PR that fixes these files so that PRs like this stay cleaner.

savsa (approved):

src/main/webapp/js/planner.js
```diff
@@ -78,7 +80,7 @@ function createTable(tableData, courseContainer) {
     const row = document.createElement('tr');
     const cell = document.createElement('td');
     const semesterLabel = document.createElement('b');
-    semesterLabel.innerText = `Semester ${i + 1}:`;
+    semesterLabel.innerText = `Semester ${i + 1} (${creditsData[i]} Credits):`;
```
triplehair, line 87: Maybe add a check that tableData and creditsData have the same size, to make the code more robust.

triplehair (approved):

wli (approved):

PR #51, opened 2020-07-20
Add ability to delete calendar events

### LINKED ISSUE #33: Add ability to remove courses from the front end calendar

savsa:
Previously, deleting calendar events by hitting the 'trash' icon on the course list didn't do anything. Now it removes that course front the front end calendar.


src/main/webapp/js/courseSelector.js
```diff
@@ -148,27 +153,28 @@ export const CollegePlanner = (() => {
     deleteButtonElement.setAttribute(
         'class', 'float-right rounded-circle border-0');
     deleteButtonElement.addEventListener('click', () => {
+      Calendar.removeCourse(liElement.getAttribute('value'));
```
ramyabuva: this might cause an error for the planner page that also depends on this file -- is there possible way to add this event listener in the calendar.js file?

src/main/webapp/js/courseSelector.js
```diff
@@ -148,27 +153,28 @@ export const CollegePlanner = (() => {
...
+   *     success, warning, danger)
+   * @param {Element} container the container you want to display the alert in
+   */
+  function createAlert(message, type, container) {
+    const alert = document.createElement('div');
+    alert.setAttribute('class', `alert alert-${type}`);
+    alert.setAttribute('role', 'alert');
+    alert.appendChild(document.createTextNode(message));
+    container.appendChild(alert);
+  }
```
ramyabuva: merge or rebase onto master to get these diffs to go away.

ramyabuva (changes requested):

src/main/webapp/js/courseSelector.js
savsa: What would cause the error? The `Calendar` namespace?

ramyabuva: When calling that function, it would try to modify elements that don't exist in the other page

naaoli (approved):

src/main/webapp/js/calendar.js
```diff
@@ -49,8 +49,16 @@ const enumDays = {
...
 
+/**
+ * Maps a course id to the array of the corresponding schedule ids on the
+ * calendar. Note that a single course with 3 meeting days has 3 different
+ * schedule ids. This is why we need an array to house them.
+ * @type {{courseId: string}, {scheduleIds: array}}
+ */
+let scheduleInfo = {};
+
 export const Calendar = (() => {
```
wli: no need for anonymous function here too. let's move this to the bottom.

src/main/webapp/js/calendar.js
```diff
@@ -167,7 +175,25 @@ function createSchedule(course, startDate, endDate) {
     isReadOnly: true,
   }]);
   // Each schedule needs a unique ID. We'll start at 1 and go up from there.
-  id++;
+  if (scheduleInfo[course.course_id] == null) {
```
wli: javascript has a weird concept of null vs undefined. my console actually shows `undefined` as the value that this would return. a safer was is just `course.course_id in scheduleInfo`

src/main/webapp/js/calendar.js
```diff
@@ -167,7 +175,25 @@ function createSchedule(course, startDate, endDate) {
...
+  }
+  scheduleInfo[course.course_id].push(id++);
+}
+
+/**
+ * Removes a schedule on the calendar using the toast API.
+ * @param {string} course_id The course_id of the course we want to remove.
+ */
+function removeCourse(course_id) {
+  const scheduleIds = scheduleInfo[course_id];
```
wli: same here

wli (changes requested):

PR #52, opened 2020-07-20
Prettify files

ramyabuva:
Prettifies files that were not prettified before previous pull requests.


savsa (approved):

wli (approved):
Do we know why the files weren't prettified in previous PRs?

savsa:
> Do we know why the files weren't prettified in previous PRs?

@wli It's possible that we prettify the files for the original pushes but then forget to prettify it again after we've made some changes to address the comments. Or we forget to do it one last time before merging. Setting up automatic formatting when pushing would be interesting.

PR #62, opened 2020-07-23
Have randomly selected schedule colors

### LINKED ISSUE #31: Have each calendar event be a different color
Currently each event/course/schedule added to the calendar on the front end is hard coded to be the same color. Make it so that each event is (randomly?) assigned a different color than the others.

savsa:
All courses on the calendar will get randomized colors from a selected list. Previously, all courses were defaulted to be light blue.

Screenshot: https://screenshot.googleplex.com/QTEh8zEs0q1.png

Closes #31.


src/main/webapp/js/lib/calendar.js
```diff
@@ -49,6 +47,34 @@ const enumDays = {
   DATE_SATURDAY: 6,
 };
 
+/**
+ * The array of colors that a calendar schedule can have.
+ * @type {array}
```
wli: `@type {Array.<string>}`

src/main/webapp/js/lib/calendar.js
```diff
@@ -49,6 +47,34 @@ const enumDays = {
...
+  '#491f87',
+  '#8f1088',
+];
+
+/**
+ * A duplicated version of the scheduleColors array. It's used to 'reset' the
+ * scheduleColors array when a new course schedule is requested.
+ * @type {array}
+ */
+const originalColors = [...scheduleColors];
```
wli: top-level constants like this should be declared like `ORIGINAL_COLORS`.

https://google.github.io/styleguide/jsguide.html#naming-constant-names

wli: i think it would make more sense to have `const ORIGINAL_COLORS = ['#ff5f4a', ...]`, and reset `scheduleColors` when it makes sense below. This is kind of "inverting" which one is the constant since you're pulling values from a non-const.

src/main/webapp/js/lib/calendar.js
```diff
@@ -206,6 +243,19 @@ function createDateFromTimeString(time, day) {
...
+  scheduleColors = [...originalColors];
+  randomizeColors();
+}
+
+/**
+ * Randomizes the scheduleColors array.
+ */
+function randomizeColors() {
+  scheduleColors.sort(() => {
+    return 0.5 - Math.random();
```
wli: Highly recommend you read this article https://medium.com/@nitinpatel_20236/how-to-shuffle-correctly-shuffle-an-array-in-javascript-15ea3f84bfb (and the [supporting analysis article](http://www.robweir.com/blog/2010/02/microsoft-random-browser-ballot.html)) for why this is not a good approach.

wli (changes requested):

src/main/webapp/js/lib/calendar.js
```diff
@@ -146,14 +181,16 @@ function addCourseToCalendar(course, startTime, endTime, day) {
  * @param {DateTime} startDate The DateTime object of the start time.
  * @param {DateTime} endDate The DateTime object of the end time.
  */
-function createSchedule(course, startDate, endDate) {
-  // TODO(savsa): Have each schedule be a different color. The color is
-  // currently hard coded to be blue so all schedules look the same.
+function createSchedule(course, startDate, endDate, color) {
+  // The default color is hard-coded to be blue;
+  if (color == null) {
+    color = '#00a9ff';
```
wli: this should be listed as a constant near the top of the file: `const DEFAULT_SCHEDULE_COLOR = '#00a9ff';`

wli (changes requested):

src/main/webapp/js/lib/calendar.js
```diff
@@ -206,6 +249,22 @@ function createDateFromTimeString(time, day) {
...
+ * Randomizes the scheduleColors array.
+ */
+function randomizeColors() {
+  for (let i = scheduleColors.length - 1; i > 0; i--) {
+    const j = Math.floor(Math.random() * i);
+    const temp = scheduleColors[i];
+    scheduleColors[i] = scheduleColors[j];
+    scheduleColors[j] = temp;
+  }
 }
```
ramyabuva: it may be better to not randomize this array -- because we have pagination now, it may be better to stay consistent with colors between schedules (meaning a consistent order before and after clearing the schedule).

savsa: Maybe randomize it once and then never again? Otherwise every schedule would be rainbow colored starting from red, orange, yellow, etc.

ramyabuva: I think that's a good idea

wli (approved):

triplehair (approved):

PR #66, opened 2020-07-24
Adds pagination for when multiple schedules are returned

### LINKED ISSUE #64: Create pagination for multiple returned schedules

ramyabuva:
https://screenshot.googleplex.com/iapGVsytApT.png

This creates buttons for users to click through different schedules (if the servlet returns multiple viable schedules). 

Closes #64


src/main/webapp/js/pages/index.js
```diff
@@ -34,10 +34,40 @@ document.querySelector('.course-list').addEventListener('click', () => {
...
+    }
+    pageList.appendChild(nextPage);
+    const scheduleNum = schedules[i - 1];
+    nextPage.addEventListener('click', () => {
+      Calendar.clear();
+      // display the correct schedule depending on the button clicked
+      selected.forEach(course => {
+        Calendar.addCourse(courseInfo[course], scheduleNum[course]);
+      });
+      const a_elements = pageList.getElementsByTagName('a');
```
savsa: should use camel case

savsa (approved):

src/main/webapp/js/pages/index.js
```diff
@@ -34,10 +34,40 @@ document.querySelector('.course-list').addEventListener('click', () => {
...
     Calendar.addCourse(courseInfo[course], schedules[0][course]);
   });
+
+  // if more than 1 schedule returned, create pagination
+  if (schedules.length == 1) {
+    return;
+  }
+  const pageList = document.getElementById('calendar-pagination');
+  pageList.innerText = ''; // clear any existing children from the element
+  for (let i = 1; i <= schedules.length; i++) {
```
wli: let's also do `i <= schedules.length && i <= MAX_PAGINATION_SCHEDULES` (new constant defined at the top)?

src/main/webapp/js/pages/index.js
```diff
@@ -34,10 +34,40 @@ document.querySelector('.course-list').addEventListener('click', () => {
...
+    const nextPage = document.createElement('a');
+    nextPage.innerText = i;
+    // by default, set first page to active
+    if (i == 1) {
+      nextPage.setAttribute('class', 'active');
+    }
+    pageList.appendChild(nextPage);
+    const scheduleNum = schedules[i - 1];
+    nextPage.addEventListener('click', () => {
+      Calendar.clear();
```
wli: we should pull this out into a common function so you can don't duplicate the code here (the clear + addCourse part) with doing it above.

src/main/webapp/js/pages/index.js
```diff
@@ -34,10 +34,40 @@ document.querySelector('.course-list').addEventListener('click', () => {
...
+      // display the correct schedule depending on the button clicked
+      selected.forEach(course => {
+        Calendar.addCourse(courseInfo[course], scheduleNum[course]);
+      });
+      const aElements = pageList.getElementsByTagName('a');
+      // remove active label from any other child element
+      for (var a = 0; a < aElements.length; a++) {
+        aElements[a].setAttribute('class', '');
```
wli: instead of clearing them all, `aElements[a].classList.remove('active');` would be better, in case anyone needs to add additional classes later

src/main/webapp/js/pages/index.js
```diff
@@ -34,10 +34,40 @@ document.querySelector('.course-list').addEventListener('click', () => {
...
+      }
+      // set current page to active
+      nextPage.setAttribute('class', 'active');
```
wli: same here with adding a class

src/main/webapp/js/pages/index.js
```diff
@@ -34,10 +34,40 @@ document.querySelector('.course-list').addEventListener('click', () => {
...
+    const scheduleNum = schedules[i - 1];
+    nextPage.addEventListener('click', () => {
+      Calendar.clear();
+      // display the correct schedule depending on the button clicked
+      selected.forEach(course => {
+        Calendar.addCourse(courseInfo[course], scheduleNum[course]);
+      });
+      const aElements = pageList.getElementsByTagName('a');
+      // remove active label from any other child element
+      for (var a = 0; a < aElements.length; a++) {
```
wli: it looks like you don't actually need the `a` variable. It would be cleaner to just do:

```
for (const element of aElements)
```

src/main/webapp/style.css
```diff
@@ -6,3 +6,23 @@
   color: #858796;
   font-size: 15px;
 }
+
+.pagination {
+  display: inline-block;
```
wli: why do you need to override all of this? by default, I think `.pagination` applies a `display: flex`, and you wouldn't need to `float: left` below. I think the key is that you need to add the `page-item` class to each item. (Also, probably worth it to swap the outer element to a `<ul>`

https://getbootstrap.com/docs/4.0/components/pagination/

wli (changes requested):

triplehair (approved):

src/main/webapp/js/pages/index.js
```diff
@@ -21,7 +21,7 @@ window.addEventListener('load', () => {
 });
 
 document.querySelector('.course-list').addEventListener('click', () => {
-  Calendar.clear();
+  const MAX_PAGINATION_SCHEDULES = 15;
```
wli: typically these constant declarations should go at the top of the file, outside of any function (line 17).

src/main/webapp/js/pages/index.js
```diff
@@ -34,10 +34,41 @@ document.querySelector('.course-list').addEventListener('click', () => {
       course => {sections2[course] = courseInfo[course].sections[1]});
   const schedules = [sections1, sections2];
 
-  // TODO(ramyabuva): add pagination for multiple schedules
-  selected.forEach(course => {
-    Calendar.addCourse(courseInfo[course], schedules[0][course]);
-  });
+  function addScheduleToCalendar(schedule){
```
wli: instead of nesting this function declaration, can this go outside as a helper function?

src/main/webapp/js/pages/index.js
```diff
@@ -34,10 +34,41 @@ document.querySelector('.course-list').addEventListener('click', () => {
...
+    Calendar.clear();
+    selected.forEach(course => {
+      Calendar.addCourse(courseInfo[course], schedule[course]);
+    });
+  }
+
+  addScheduleToCalendar(schedules[0]);
```
wli: why are you selecting only the first element? if it's just for testing (until we hook up other parts of the app), please make a comment explaining.

src/main/webapp/js/pages/index.js
```diff
@@ -34,10 +34,41 @@ document.querySelector('.course-list').addEventListener('click', () => {
...
+  pageList.innerText = ''; // clear any existing children from the element
+  for (let i = 1; i <= schedules.length && i <= MAX_PAGINATION_SCHEDULES; i++) {
+    const nextPage = document.createElement('li');
+    nextPage.innerText = i;
+    // by default, set first page to active
+    if (i == 1) {
+      nextPage.setAttribute('class', 'active');
+    }
+    pageList.appendChild(nextPage);
+    const NumSchedule = schedules[i - 1];
```
wli: `NumSchedule` -> `numSchedule`. variables declared in functions start with a lowercase. But also, is `schedule` a better name for this? Why is "num" relevant?

src/main/webapp/style.css
```diff
@@ -6,3 +6,17 @@
...
+}
+
+.pagination li.active {
+  background-color: #3d64d3;
+  color: white;
+}
+
+.pagination li:hover:not(.active) {
+  background-color: #ddd;
+}
```
wli: nit: missing newline

wli (changes requested):

src/main/webapp/js/pages/index.js
ramyabuva: As it is, it can't go outside of the function because it uses variables inside the scope of the function. I could add those variables as parameters instead and move the nested function outside?

wli: Yep, that sounds reasonable.

src/main/webapp/js/pages/index.js
```diff
@@ -15,13 +15,28 @@
...
   Calendar.initCalendar();
   CollegePlanner.getDepartmentOptions();
 });
 
-document.querySelector('.course-list').addEventListener('click', () => {
+/**
+ * Renders the specified schedule on the calendar
+ * @param {Object} schedule JSON object mapping course to the section id
+ * @param {Object} courseInfo JSON object mapping course to course details
+ * @param {Object} selected Array containing all of the course_ids selected by
```
wli: I think this should be `@param {Array.<string>}`

src/main/webapp/js/pages/index.js
```diff
@@ -34,10 +49,35 @@ document.querySelector('.course-list').addEventListener('click', () => {
       course => {sections2[course] = courseInfo[course].sections[1]});
   const schedules = [sections1, sections2];
 
-  // TODO(ramyabuva): add pagination for multiple schedules
-  selected.forEach(course => {
-    Calendar.addCourse(courseInfo[course], schedules[0][course]);
-  });
+  // by default, add the first schedule to the calendar after response received
```
wli: This is one of the smallest nits possible, but general style guide rules are that comments should usually be proper English sentences.

Specifically: capitalize the start of your comments, and end them with a period. Same for other comments below.

wli (approved):

PR #70, opened 2020-07-24
Create Semester Scheduler

### LINKED ISSUE #27: Create Semester Scheduling Algorithm and necessary classes

naaoli:


triplehair (approved):

src/main/java/com/google/collegeplanner/data/Schedule.java
```diff
@@ -24,6 +24,13 @@ public Schedule() {
     sections = new ArrayList<Section>();
   }
 
+  public Schedule(Schedule copy) {
+    sections = new ArrayList<Section>();
```
wli: I think either `new ArrayList<>(copy.getSections());` or `Collections.copy(copy.getSections(), sections)` is preferred over manually iterating

src/main/java/com/google/collegeplanner/data/Schedule.java
```diff
@@ -24,6 +24,13 @@ public Schedule() {
     sections = new ArrayList<Section>();
   }
 
+  public Schedule(Schedule copy) {
```
wli: I think naming this `Schedule schedule` would be clearer.

wli (changes requested):

ramyabuva (approved):

src/main/java/com/google/collegeplanner/data/Schedule.java
```diff
@@ -0,0 +1,105 @@
...
+public class Schedule {
+  private ArrayList<Section> sections;
+
+  public Schedule() {
+    sections = new ArrayList<Section>();
+  }
+
+  public Schedule(Schedule schedule) {
+    sections = new ArrayList<Section>();
+    Collections.copy(schedule.getSections(), sections)
```
wli: I think Collections needs to be imported - does this pass tests?

src/test/java/com/google/collegeplanner/SemesterSchedulerTest.java
```diff
@@ -0,0 +1,211 @@
...
+
+    assertEqualsLists(expectedSchedules, scheduler.getPossibleSchedules());
+  }
+
+  /**
+   * This function checks if two lists are the same (order does not matter)
+   * @param expected the list of expected schedules
+   * @param actual the list of actual schedules
+   */
+  public void assertEqualsLists(ArrayList<Schedule> expected, ArrayList<Schedule> actual) {
```
wli: any chance we could use `containsExactlyElementsIn` instead of writing your own (or using some other open source alternative)? This is part of the `com.google.common.truth` package - though I'm not sure how difficult it is to pull in (https://github.com/google/truth).

wli (changes requested):

savsa (approved):

src/test/java/com/google/collegeplanner/SemesterSchedulerTest.java
```diff
@@ -0,0 +1,196 @@
...
+    expectedSchedules.get(1).addClass(mathClasses.get(1));
+    expectedSchedules.get(1).addClass(chemClasses.get(0));
+
+    expectedSchedules.get(2).addClass(compSciClasses.get(1));
+    expectedSchedules.get(2).addClass(englishClasses.get(0));
+    expectedSchedules.get(2).addClass(mathClasses.get(0));
+    expectedSchedules.get(2).addClass(chemClasses.get(0));
+
+    ArrayList<Schedule> actualSchedules = scheduler.getPossibleSchedules();
+    Assert.assertTrue(expectedSchedules.containsAll(actualSchedules) &&
```
wli: if you ever finding yourself using a `&&` for test assertions, you can almost always split this up to help yourself debug easier in the future. When an assertion fails, you want it to be as easy to find out what is wrong as possible.

```
Assert.assertTrue(expectedSchedules.containsAll(actualSchedules))
Assert.assertEquals(expectedSchedules.size(), actualSchedules.size());
```

In addition, while `.containsAll` will tell you that something is wrong, it won't tell you which element is missing, which is important in debugging. Here's an example where you shouldn't need to do any copying: https://github.com/googleinterns/InterviewMe/blob/e3df6cf13c3d7e256cfb56f2a4d1e251156c9972/src/test/java/com/google/sps/data/ScheduledInterviewServletTest.java#L219

wli (changes requested):

src/main/java/com/google/collegeplanner/data/Schedule.java
```diff
@@ -61,6 +64,10 @@ private boolean conflictsWithSchedule(Section section) {
     return sections;
   }
 
+  public void removeLastClass() {
+    sections.remove(sections.size() - 1);
```
wli: This will crash if you remove when `sections.size() == 0`. Also, adding a quick test for this would be helpful to ensure this doesn't fail in unexpected ways.

src/main/java/com/google/collegeplanner/data/SemesterScheduler.java
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  /**
+   * This function is a recursivly nested for loop.
+   * @param sectionListSizes the end case number for the nested for loops
+   * @param level the depth of the nested for loop the method is currently on.
+   */
+  private void nestedLoop(int[] sectionListSizes, int level) {
+    if (possibleSchedules.size() >= MAX_SCHEDULES) {
+      System.out.println("too big");
```
wli: We try to not leave logging statements like this in the code. If we want verbose logging, we can pull in a framework to do so, but let's leave that for a future PR. For now, can you just remove this line?

src/main/java/com/google/collegeplanner/data/Schedule.java
```diff
@@ -61,6 +64,10 @@ private boolean conflictsWithSchedule(Section section) {
     return sections;
   }
 
+  public void removeLastClass() {
```
wli, line 70: Missing documentation.

wli (changes requested):

wli (approved):

PR #74, opened 2020-07-28
Download data from the UMD API to Datastore

### LINKED ISSUE #75: Store Course data in datastore

savsa:
Create `DatastoreServlet.java` and add tests. The datastore servlet downloads course data from the UMD API and stores it in datastore.

I removed the `gradingSystem` variable from `Course.java` because I don't think it's that important to store anymore, on top of the fact that it adds a little bit of complexity storing it because it's an array.

Note: this datastore feature isn't fully finished yet. I will have at least one more upcoming PR to finish up the rest.

Closes #75.


ramyabuva (approved):

src/main/java/com/google/collegeplanner/data/Course.java
```diff
@@ -101,14 +96,33 @@ public Course(String courseId, String name, String semester, int credits, String
...
-        (int) json.get("credits"), (String) json.get("dept_id"), (String) json.get("description"),
-        (String[]) ((JSONArray) json.get("grading_method")).toArray(),
-        (String) ((JSONObject) json.get("relationships")).get("coreqs"),
-        (String) ((JSONObject) json.get("relationships")).get("prereqs"),
-        (String) ((JSONObject) json.get("relationships")).get("restrictions"),
-        (String) ((JSONObject) json.get("relationships")).get("additional_info"),
-        (String) ((JSONObject) json.get("relationships")).get("credit_granted_for"));
+    try {
+      this.credits = Integer.parseInt((String) json.get("credits"));
+    } catch (Exception e) {
```
wli: catch specific exceptions here - probably `NumberFormatException`

src/main/java/com/google/collegeplanner/data/Course.java
```diff
@@ -101,14 +96,33 @@ public Course(String courseId, String name, String semester, int credits, String
...
+      this.credits = 0;
+    }
+
+    this.courseId = (String) json.get("course_id");
```
wli, line 107: do any of these fields need to be validated? are empty fields values for any of these fields ok? or should you throw your own error if you encounter an incomplete object?

src/main/java/com/google/collegeplanner/servlets/DatastoreServlet.java
```diff
@@ -0,0 +1,145 @@
...
+      JSONArray coursesArray = apiUtil.getJsonArray(uri);
+      if (coursesArray == null) {
+        respondWithError(HttpServletResponse.SC_INTERNAL_SERVER_ERROR, response);
+        return;
+      } else if (coursesArray.size() == 0) {
+        respondWithError(HttpServletResponse.SC_OK, response);
+        return;
+      }
+      addCourses(coursesArray, response);
+    } while (true);
```
wli: instead of an actual infinite loop, I think you should set an upper limit (100 pages? 1000?) just in case. This should be declared as a constant.

This also helps protect against a case where the API returns the same thing no matter what page number you pass it.

src/main/java/com/google/collegeplanner/servlets/DatastoreServlet.java
```diff
@@ -0,0 +1,145 @@
...
+      } catch (URISyntaxException e) {
+        respondWithError(HttpServletResponse.SC_INTERNAL_SERVER_ERROR, response);
+        return;
+      }
+      JSONArray coursesArray = apiUtil.getJsonArray(uri);
+      if (coursesArray == null) {
+        respondWithError(HttpServletResponse.SC_INTERNAL_SERVER_ERROR, response);
+        return;
+      } else if (coursesArray.size() == 0) {
+        respondWithError(HttpServletResponse.SC_OK, response);
```
wli: why is this `respondWithError`, but it's actually an `SC_OK`?

src/main/java/com/google/collegeplanner/servlets/DatastoreServlet.java
```diff
@@ -0,0 +1,145 @@
...
+    // Loop through each page's courses.
+    for (Object jsonObject : coursesArray) {
+      JSONObject courseJson = (JSONObject) jsonObject;
+
+      // Create a Course object from the JSONObject for parsing.
+      Course course = new Course(courseJson);
+
+      URI uri;
+      try {
+        uri = new URI("https://api.umd.io/v1/courses/" + course.getCourseId() + "/sections");
```
wli: I think you'll want to escape the courseId here for safety too.

src/main/java/com/google/collegeplanner/servlets/DatastoreServlet.java
```diff
@@ -0,0 +1,145 @@
...
+      // TODO(savsa): create Section class to handle JSON object parsing.
+      // For now, just store the section id.
+      JSONObject sectionJson = (JSONObject) jsonObject;
+      EmbeddedEntity sectionEntity = new EmbeddedEntity();
+      sectionEntity.setProperty("section_id", (String) sectionJson.get("section_id"));
+      sectionEntities.add(sectionEntity);
+    }
+
+    Entity entity = new Entity("Course");
+    entity.setProperty("course_id", course.getCourseId());
```
wli: what happens if you run this function twice? It seems like you'll get duplicated entities.

src/test/java/com/google/collegeplanner/DatastoreTest.java
```diff
@@ -0,0 +1,180 @@
...
+  }
+
+  @After
+  public void after() {
+    helper.tearDown();
+  }
+
+  /* Run this test twice to prove we're not leaking any state across tests. */
+  @Test
+  public void testDatastoreInsertion1() {
```
wli: curious - was this approach recommended somewhere? i'm not sure this is necessary.

src/test/java/com/google/collegeplanner/DatastoreTest.java
```diff
@@ -0,0 +1,180 @@
...
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+import org.skyscreamer.jsonassert.JSONAssert;
+import org.skyscreamer.jsonassert.JSONCompareMode;
+
+/** Tests DatastoreServlet */
+@RunWith(JUnit4.class)
+public final class DatastoreTest {
```
wli: DatastoreServletTest

src/test/java/com/google/collegeplanner/DatastoreTest.java
```diff
@@ -0,0 +1,180 @@
...
+    Assert.assertEquals(1, datastore.prepare(new Query("Course")).countEntities());
+  }
+
+  private void insertionTest() {
+    Assert.assertEquals(0, datastore.prepare(new Query("Course")).countEntities());
+    datastore.put(new Entity("Course"));
+    datastore.put(new Entity("Course"));
+    Assert.assertEquals(2, datastore.prepare(new Query("Course")).countEntities());
+  }
+}
```
wli: missing a test that would do multiple pages of fetches

wli (changes requested):

src/test/java/com/google/collegeplanner/DatastoreTest.java
savsa: None of the other tests have the word 'servlet' in them

src/test/java/com/google/collegeplanner/DatastoreTest.java
savsa: It was an example they gave in the docs but I guess it's not necessary

src/main/java/com/google/collegeplanner/servlets/DatastoreServlet.java
savsa: Oh oops. Should I have it return any error code at all or have it return 200?

src/test/java/com/google/collegeplanner/DatastoreTest.java
wli: I commented on it in one of the other PRs, but it's an oversight in general. This is something we should fix. Test files for `Filename.java` should be `FilenameTest.java`. Otherwise, if you create a Datastore.java later, you're going to have a conflict here.

src/test/java/com/google/collegeplanner/DatastoreTest.java
wli: I think just having any two tests below should provide the same functional coverage of this case.

src/main/java/com/google/collegeplanner/servlets/DatastoreServlet.java
wli: This isn't an error unless the `page == 0` right? It should the main path to a 200 in most cases.

src/main/java/com/google/collegeplanner/servlets/DatastoreServlet.java
savsa: how would you prevent that? by using keys?

wli: Keys is one way. But more likely, you'd want to look up any existing records and update it with your newly pulled data.

src/main/java/com/google/collegeplanner/data/Course.java
savsa: description, prereqs, coreqs, etc are oftentimes null. I don't think those matter that much, but a null course_id for example would be bad. So we'd maybe want to skip it? Not sure

src/main/java/com/google/collegeplanner/servlets/DatastoreServlet.java
savsa: Actually I think because I'm using `datastore.put()` instead of `datastore.add()`, it overwrites the previous entry if it already existed.

wli: Which fields are you using to determine whether it "already existed"?

savsa: course_id?

src/main/java/com/google/collegeplanner/servlets/DatastoreServlet.java
```diff
@@ -0,0 +1,179 @@
...
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    // Loop through the pages.
+    // We don't know how many pages there are beforehand. We stop looping when the new page
+    // doesn't have any results.
+    int page = 1;
+    do {
+      URI uri;
+      try {
+        URIBuilder builder = new URIBuilder("https://api.umd.io/v1/courses");
+        builder.setParameter("page", Integer.toString(page));
```
triplehair, line 76: Can we say we want to fetch courses in a batch before page 200? I am thinking if we can do that we don't need the loop to call api multiple times.

src/main/java/com/google/collegeplanner/servlets/DatastoreServlet.java
```diff
@@ -0,0 +1,179 @@
...
+      } else {
+        // The entity already exists - modify it instead.
+        courseEntity = limitedResults.get(0);
+      }
+
+      URI uri;
+      try {
+        uri = new URI("https://api.umd.io/v1/courses/"
+            + URLEncoder.encode(course.getCourseId(), StandardCharsets.UTF_8.toString())
+            + "/sections");
```
triplehair, line 135: Do we have to call .../sections specifically in order to get the sections information? I am wondering if we can get that from the first api call and store it in Course?

src/main/java/com/google/collegeplanner/servlets/DatastoreServlet.java
savsa: the api only returns 30 courses max at a time so we have to cycle through the "pages" to get all of them

src/main/java/com/google/collegeplanner/servlets/DatastoreServlet.java
savsa: yes we have to call /sections. calling /course only gives us the section_ids basically, but we have to call /sections to get the full detailed info about each section

src/main/java/com/google/collegeplanner/data/Course.java
```diff
@@ -84,31 +80,52 @@
...
+  public Course(JSONObject json) throws Exception {
+    try {
+      this.credits = Integer.parseInt((String) json.get("credits"));
+    } catch (NumberFormatException e) {
+      this.credits = 0;
+    }
+
+    this.courseId = (String) json.get("course_id");
+    if (this.courseId == null) {
+      throw new Exception("Null course_id.");
```
wli: this should likely be ParseError instead of a generic Exception

src/main/java/com/google/collegeplanner/servlets/DatastoreServlet.java
```diff
@@ -0,0 +1,179 @@
...
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import org.apache.http.client.utils.URIBuilder;
+import org.json.simple.JSONArray;
+import org.json.simple.JSONObject;
+
+/** Queries the UMD API and downloads the data to datastore. */
+@WebServlet("/api/download")
+public class DatastoreServlet extends BaseServlet {
+  final int PAGE_LIMIT = 200;
```
wli, line 50: add documentation for what this applies to.

src/main/java/com/google/collegeplanner/servlets/DatastoreServlet.java
```diff
@@ -0,0 +1,179 @@
...
+   * @param coursesArray The course json from the UMD API.
+   * @param response The HttpServletResponse object.
+   */
+  private void addCourses(JSONArray coursesArray, HttpServletResponse response) throws IOException {
+    if (coursesArray == null) {
+      return;
+    }
+
+    // Loop through each page's courses.
+    for (Object jsonObject : coursesArray) {
```
wli, line 105: I'm curious - can you just do `JSONObject jsonObject : coursesArray`? Or will that cause a compile error?

src/main/java/com/google/collegeplanner/servlets/DatastoreServlet.java
```diff
@@ -0,0 +1,179 @@
...
+      Course course;
+      try {
+        course = new Course(courseJson);
+      } catch (Exception e) {
+        respondWithError(HttpServletResponse.SC_INTERNAL_SERVER_ERROR, response);
+        return;
+      }
+
+      // Check if the entity already exists in datastore, and modify it if it does.
+      Query query = new Query("Course").setFilter(
```
wli, line 118: just double checking - did you create an index on course_id (and/or is it a primary/secondary key)? otherwise, this could be a linear search over all your courses.

wli, line 118: side question - is there a way in datastore to enforce uniqueness on an attribute like `course_id`, so we'll never accidentally commit two at the same time (if there are two scripts running simultaneously for example).

wli (changes requested):

src/main/java/com/google/collegeplanner/servlets/DatastoreServlet.java
savsa: no I don't have an index yet. i'll need to look into how to make one

savsa: Enforce it with a datastore setting? I don't believe so. The way I did it queries the actual database anyways, so if one script is running and it adds the entity slightly before the other script does, then this query should pick up that the entity is already in the database, right?

savsa: https://cloud.google.com/datastore/docs/concepts/indexes

"By default, a Datastore mode database automatically predefines an index for each property of each entity kind. These single property indexes are suitable for simple types of queries."

Aren't all entity properties indexed by default though?

src/main/java/com/google/collegeplanner/servlets/DatastoreServlet.java
savsa: Yeah, it's an incompatible types error

wli: Sounds good, thanks for checking.

src/main/java/com/google/collegeplanner/servlets/DatastoreServlet.java
wli: Ah ok, good to know. Should be good then.

triplehair (approved):

src/main/java/com/google/collegeplanner/servlets/DatastoreServlet.java
wli: Concurrent writes are difficult to get right, especially in distributed systems. There's always a chance that both writes will try to happen simultaneously.

Also, the concurrent scripts likely won't hit the same server, so it's not always possible to pick up an "already-written" entity in the database due to small sync/propagation delays.

src/main/java/com/google/collegeplanner/servlets/DatastoreServlet.java
wli: It looks like datastore does "overwrite" based on the key of each object. Is `course_id` defined as the key? It doesn't look like it.

https://cloud.google.com/appengine/docs/standard/java/datastore/creating-entities#Updating_entities

src/main/java/com/google/collegeplanner/servlets/DatastoreServlet.java
savsa: Through my research online I haven't been able to find a better solution. If we needed to seriously handle lots of concurrent writes then we would've had to use Firestore or something similar.

src/main/java/com/google/collegeplanner/servlets/DatastoreServlet.java
savsa: I was slightly off about put vs add. But I changed my approach anyways. if it finds that the entity already exists, then that entity gets edited with the new properties (the key stays the same), and then it gets re `put` into the database. And since I edited the entity that was already in the database and the key is the same, then it overwrites that entity with the new modified one. If the entity doesn't already exists, it creates a new one and then `put`s it into the database.

I added an extra test that tests this and the code passes.

src/main/java/com/google/collegeplanner/servlets/DatastoreServlet.java
wli: It looks like we can solve this by just using the course_id as the key. You would trade off correctness with any potential key distribution problems - seems like a worthwhile tradeoff for our use case here to me.

naaoli (approved):

src/main/java/com/google/collegeplanner/data/Course.java
```diff
@@ -84,31 +80,52 @@
   private String creditGrantedFor;
 
   public Course(String courseId, String name, String semester, int credits, String departmentId,
-      String description, String[] gradingMethod, String corequisites, String prerequisites,
-      String restrictions, String additionalInfo, String creditGrantedFor) {
+      String description, String corequisites, String prerequisites, String restrictions,
+      String additionalInfo, String creditGrantedFor) {
     this.courseId = courseId;
```
wli, line 85: since courseId is being used as a key now, do we need to normalize it (uppercase, lowercase, etc) before putting it in? Otherwise MATH101 and Math101 will be two separate objects.

wli (changes requested):
Just one last small thing to double check.

src/main/java/com/google/collegeplanner/data/Course.java
```diff
@@ -84,31 +80,54 @@
...
```
wli, line 85: you would want to do it here too

wli (approved):

PR #76, opened 2020-07-28
Add save and sign in modal

ramyabuva:
Adds modals that allow the user to name and save their plans depending on whether they are logged in or not. 
https://screenshot.googleplex.com/fzY0fywywPP.png

### DIFF
--- a/src/main/webapp/js/pages/planner.js
+++ b/src/main/webapp/js/pages/planner.js
@@ -28,9 +28,15 @@ async function getPlan() {
     selectedClasses: selectedClasses,
     semesters: document.getElementById('semesters').value
   };
-  courseContainer.innerText = '';  // Clears card body to get rid of spinner
+  courseContainer.innerText = '';  // Clears card body to get rid of spinner.
   let response;
   let courseList;
+  // Clear save button from header.
+  document.getElementById('plan-header').innerText = '';
+  // Replace form to clear all event listeners so it doesn't save old schedules.
+  const oldNameForm = document.getElementById('plan-name');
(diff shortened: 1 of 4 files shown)


src/main/webapp/js/pages/planner.js
```diff
@@ -59,6 +69,42 @@ async function getPlan() {
   }
 }
 
+function savePlan(courseList) {
```
savsa, line 77: needs comment

src/main/webapp/js/pages/planner.js
```diff
@@ -59,6 +69,42 @@ async function getPlan() {
...
+  // TODO(ramyabuva): Send tokens to servlet to save plans in Datastore.
```
savsa: needs issue?

savsa (approved):

src/main/webapp/js/pages/planner.js
```diff
@@ -28,9 +28,15 @@ async function getPlan() {
     selectedClasses: selectedClasses,
     semesters: document.getElementById('semesters').value
   };
-  courseContainer.innerText = '';  // Clears card body to get rid of spinner
+  courseContainer.innerText = '';  // Clears card body to get rid of spinner.
   let response;
   let courseList;
+  // Clear save button from header.
+  document.getElementById('plan-header').innerText = '';
+  // Replace form to clear all event listeners so it doesn't save old schedules.
```
wli, line 36: not the cleanest approach (you could clear the event listeners manually instead), but in the interest of time, this will work.

wli (approved):

triplehair (approved):

PR #78, opened 2020-07-30
Verify tokens on back end and save plans to Datastore

### LINKED ISSUE #77: Send id tokens to servlet to save plans in Datastore

ramyabuva:
I created a new SavePlanServlet.java to allow users to save their schedules into Datastore. To save, planner.js sends a POST request that verifies that the user authentication token is valid, and if so, saves the desired plan into Datastore. The saved plans can then be retrieved with a GET request, that also verifies the authentication token and returns the proper entities from Datastore.

I also moved getBody(), a function that gets the body of a POST request, into BaseServlet, so it can be used for all servlets.

closes #77

### DIFF
--- a/src/main/java/com/google/collegeplanner/data/Plan.java
+++ b/src/main/java/com/google/collegeplanner/data/Plan.java
@@ -14,16 +14,16 @@
 
 package com.google.collegeplanner.data;
 
+import org.json.simple.JSONObject;
+
 public class Plan {
   private final long id;
-  private final String plan;
+  private final JSONObject plan;
   private final String planName;
-  private final String timestamp;
 
-  public Plan(long id, String plan, String planName, String timestamp) {
+  public Plan(long id, JSONObject plan, String planName) {
     this.id = id;
     this.plan = plan;
     this.planName = planName;
-    this.timestamp = timestamp;
   }
 }
--- a/src/main/java/com/google/collegeplanner/data/Plan.java
+++ b/src/main/java/com/google/collegeplanner/data/Plan.java
@@ -16,6 +16,7 @@ package com.google.collegeplanner.data;
 
 import org.json.simple.JSONObject;
 
+/** This class represents a multi-semester plan. */
 public class Plan {
   private final long id;
   private final JSONObject plan;
--- a/src/main/java/com/google/collegeplanner/data/Plan.java
+++ b/src/main/java/com/google/collegeplanner/data/Plan.java
@@ -18,10 +18,12 @@ import org.json.simple.JSONObject;
 
 /** This class represents a multi-semester plan. */
 public class Plan {
-  private final long id; // The id of the Plan in Datastore.
-  private final JSONObject
-      plan; // The JSONObject that contains the plan details (courses and credits).
-  private final String planName; // The name the user specified for the plan
+  // The id of the Plan in Datastore.
+  private final long id;
+  // The JSONObject that contains the plan details (courses and credits).
+  private final JSONObject plan;
+  // The name the user specified for the plan
+  private final String planName;
 
   public Plan(long id, JSONObject plan, String planName) {
     this.id = id;
--- a/src/main/java/com/google/collegeplanner/data/Plan.java
+++ b/src/main/java/com/google/collegeplanner/data/Plan.java
@@ -18,9 +18,10 @@ import org.json.simple.JSONObject;
 
 /** This class represents a multi-semester plan. */
 public class Plan {
-  private final long id;
-  private final JSONObject plan;
-  private final String planName;
+  private final long id; // The id of the Plan in Datastore.
+  private final JSONObject
+      plan; // The JSONObject that contains the plan details (courses and credits).
+  private final String planName; // The name the user specified for the plan
 
   public Plan(long id, JSONObject plan, String planName) {
     this.id = id;
--- a/src/main/java/com/google/collegeplanner/data/Plan.java
+++ b/src/main/java/com/google/collegeplanner/data/Plan.java
@@ -1,0 +1,29 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.package com.google.collegeplanner.data;
+
+package com.google.collegeplanner.data;
+
+public class Plan {
+  private final long id;
+  private final String plan;
+  private final String planName;
+  private final String timestamp;
+
+  public Plan(long id, String plan, String planName, String timestamp) {
+    this.id = id;
+    this.plan = plan;
+    this.planName = planName;
+    this.timestamp = timestamp;
+  }
+}
--- a/src/main/java/com/google/collegeplanner/servlets/BaseServlet.java
+++ b/src/main/java/com/google/collegeplanner/servlets/BaseServlet.java
@@ -90,7 +90,7 @@ public abstract class BaseServlet extends HttpServlet {
   /**
    * Gets the JSON Representation of the body of the POST request
    */
-  public JSONObject getBody(HttpServletRequest request)
+  public JSONObject getPostRequestBody(HttpServletRequest request)
       throws IOException, ParseException, NullPointerException {
     String strBody =
         request.getReader().lines().collect(Collectors.joining(System.lineSeparator()));
--- a/src/main/java/com/google/collegeplanner/servlets/BaseServlet.java
+++ b/src/main/java/com/google/collegeplanner/servlets/BaseServlet.java
@@ -16,9 +16,13 @@ package com.google.collegeplanner.servlets;
 
 import com.google.gson.Gson;
 import java.io.IOException;
+import java.util.stream.Collectors;
 import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import org.json.simple.JSONObject;
+import org.json.simple.parser.JSONParser;
+import org.json.simple.parser.ParseException;
 
 /** Provides an interface for shared logic between the servlets. */
 public abstract class BaseServlet extends HttpServlet {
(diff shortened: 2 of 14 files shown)


src/main/java/com/google/collegeplanner/data/Plan.java
```diff
@@ -0,0 +1,29 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.package com.google.collegeplanner.data;
+
+package com.google.collegeplanner.data;
+
+import org.json.simple.JSONObject;
+
+public class Plan {
```
wli, line 20: add some documentation. What is a "Plan"?

src/main/java/com/google/collegeplanner/servlets/SavePlanServlet.java
```diff
@@ -0,0 +1,161 @@
...
+@WebServlet("/api/planner/save")
+public class SavePlanServlet extends BaseServlet {
+  GoogleIdTokenVerifier verifier;
+  DatastoreService datastore;
+
+  public SavePlanServlet() {
+    this(DatastoreServiceFactory.getDatastoreService(),
+        new GoogleIdTokenVerifier.Builder(UrlFetchTransport.getDefaultInstance(), new GsonFactory())
+            .setAudience(Collections.singletonList(
+                "267429534228-vvsi2uldmpji3rgs1qd3a41rceciaaaq.apps.googleusercontent.com"))
```
wli: this should be a `final` constant, ideally with some documentation so it's clear what this is and where you got it from.

src/main/java/com/google/collegeplanner/servlets/SavePlanServlet.java
```diff
@@ -0,0 +1,161 @@
...
+    GoogleIdToken idToken;
+    try {
+      idToken = verifier.verify(request.getParameter("idToken"));
+    } catch (GeneralSecurityException e) {
+      respondWithError(HttpServletResponse.SC_INTERNAL_SERVER_ERROR, response);
+      return;
+    }
+
+    if (idToken == null) {
+      respondWithError("Invalid user.", HttpServletResponse.SC_BAD_REQUEST, response);
```
wli: I think this should be a 401 error - what do you think?

src/main/java/com/google/collegeplanner/servlets/SavePlanServlet.java
```diff
@@ -0,0 +1,161 @@
...
+      return;
+    }
+
+    Payload payload = idToken.getPayload();
+    Query query =
+        new Query("Plan")
+            .setFilter(new FilterPredicate("user", FilterOperator.EQUAL, payload.getEmail()))
```
wli: does this need a composite index on `user, timestamp`? I'm not sure.

wli: If you're using emails as a filtering mechanism, do emails need to be normalized (lowercase, etc)?

src/main/java/com/google/collegeplanner/servlets/SavePlanServlet.java
```diff
@@ -0,0 +1,161 @@
...
+            .addSort("timestamp", SortDirection.DESCENDING);
+    List<Entity> results = datastore.prepare(query).asList(FetchOptions.Builder.withDefaults());
+
+    ArrayList<Plan> plans = new ArrayList<>();
```
wli, line 120: the `doGet` function is a bit long. Can be break this out into a `getPlans()` helper function that does some of the parsing you have in this scetion?

src/main/java/com/google/collegeplanner/servlets/SavePlanServlet.java
```diff
@@ -0,0 +1,161 @@
...
+    JSONParser parser = new JSONParser();
+    for (Entity entity : results) {
+      try {
+        long id = entity.getKey().getId();
+        JSONObject plan = (JSONObject) parser.parse((String) entity.getProperty("plan"));
+        String planName = (String) entity.getProperty("planName");
+        plans.add(new Plan(id, plan, planName));
+      } catch (ParseException e) {
+        respondWithError(
+            "Returned plans are invalid.", HttpServletResponse.SC_BAD_REQUEST, response);
```
wli: this is more of a 500 internal server error. Bad request typically implies that the client could change parameters/conditions to resolve the issue, but this is a backend data problem.

src/main/java/com/google/collegeplanner/servlets/SavePlanServlet.java
```diff
@@ -0,0 +1,161 @@
...
+   * @param plan The plan that was saved.
+   * @param planName the name the user tried to save their plan with.
+   */
+  private void addToDatastore(String email, String plan, String planName) {
+    long timestamp = System.currentTimeMillis();
+    Entity planEntity = new Entity("Plan");
+    planEntity.setProperty("plan", plan);
+    planEntity.setProperty("planName", planName);
+    planEntity.setProperty("timestamp", timestamp);
+    planEntity.setProperty("user", email);
```
wli: same here - does email need to be normalized since you're using it as a lookup mechanism?

src/main/java/com/google/collegeplanner/servlets/SavePlanServlet.java
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    Payload payload = idToken.getPayload();
+    addToDatastore(payload.getEmail(), plan, planName);
+  }
+
+  /**
+   * Writes the saved plan to datastore.
+   * @param email The email associated with the plan.
+   * @param plan The plan that was saved.
```
wli: It would be good to call out what type/data "plan" is. I'm a bit confused reading this if it's stringified json, an array, or something else (at least, without reading the frontend code).

src/main/java/com/google/collegeplanner/servlets/BaseServlet.java
```diff
@@ -82,4 +86,15 @@ public void respondWithError(String message, int status, HttpServletResponse res
     response.setContentType("application/json;");
     response.getWriter().println(new Gson().toJson(jsonObject));
   }
+
+  /**
+   * Gets the JSON Representation of the body of the POST request
+   */
+  public JSONObject getBody(HttpServletRequest request)
```
triplehair: Might be better to name it getPostRequestBody to indicate that it is only for POST request

src/main/webapp/js/pages/planner.js
```diff
@@ -74,9 +74,27 @@ async function getPlan() {
...
+    planName: document.getElementById('save-plan-as').value
+  };
+  try {
+    await fetch('/api/planner/save', {
+      method: 'POST',
+      body: JSON.stringify(data),
+    });
+    $('#savePlanModal').modal('hide');
+  } catch (err) {
+    return;
```
wli, line 99: no user feedback?

wli (changes requested):

src/main/java/com/google/collegeplanner/servlets/SavePlanServlet.java
ramyabuva: I don't think that its necessary for our purposes. Datastore has composite indexing, but it can get a little complex
([https://cloud.google.com/datastore/docs/concepts/indexes](https://cloud.google.com/datastore/docs/concepts/indexes)). I'm not sure I see a real benefit from implementing it for this, though.

savsa (approved):

src/main/java/com/google/collegeplanner/data/Plan.java
```diff
@@ -0,0 +1,30 @@
...
+// limitations under the License.package com.google.collegeplanner.data;
+
+package com.google.collegeplanner.data;
+
+import org.json.simple.JSONObject;
+
+/** This class represents a multi-semester plan. */
+public class Plan {
+  private final long id;
+  private final JSONObject plan;
```
wli, line 24: some documentation on the fields of "plan" would be really helpful.

src/main/java/com/google/collegeplanner/servlets/SavePlanServlet.java
wli: still missing i think? i'm not sure your code pushed successfully.

wli (changes requested):

triplehair (approved):

src/main/java/com/google/collegeplanner/servlets/SavePlanServlet.java
ramyabuva: it's in the diff when I look at it in "Files Changed"?

src/main/java/com/google/collegeplanner/data/Plan.java
```diff
@@ -0,0 +1,31 @@
...
+// limitations under the License.package com.google.collegeplanner.data;
+
+package com.google.collegeplanner.data;
+
+import org.json.simple.JSONObject;
+
+/** This class represents a multi-semester plan. */
+public class Plan {
+  private final long id; // The id of the Plan in Datastore.
+  private final JSONObject
```
wli: small nit, but you should reformat this as:

```
// The JSONObject that contains the plan details (courses and credits).
private final JSONObject plan;
```

It's much better to have a line break between the comment and the declaration, instead of breaking up the declaration line.

src/main/java/com/google/collegeplanner/data/Plan.java
```diff
@@ -0,0 +1,31 @@
...
+      plan; // The JSONObject that contains the plan details (courses and credits).
+  private final String planName; // The name the user specified for the plan
```
wli: style guide is two spaces after the end of a line before a comment:

```
private final String planName;  // The name the user specified for the plan
```

Note that you can optionally also move the comment to the previous line as mentioned in my comment above.

src/main/java/com/google/collegeplanner/servlets/SavePlanServlet.java
```diff
@@ -0,0 +1,183 @@
...
+      return;
+    }
+
+    if (idToken == null) {
+      respondWithError("Invalid user.", HttpServletResponse.SC_UNAUTHORIZED, response);
+      return;
+    }
+    ArrayList<Plan> plans = getPlans(idToken, response);
+    if (plans == null) {
+      return; // Error has already been written.
```
wli: Instead of hoping that the error has already been written, I would recommend writing the error here, or throwing an error in getPlans instead. It makes it easier to enforce that the proper responses are being written to the client if it generally all happens in one place.

also, need two spaces before comment

src/test/java/com/google/collegeplanner/SavePlanTest.java
```diff
@@ -0,0 +1,202 @@
...
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+import org.skyscreamer.jsonassert.JSONAssert;
+import org.skyscreamer.jsonassert.JSONCompareMode;
+
+/** Test SavePlanServlet */
+@RunWith(JUnit4.class)
+public final class SavePlanTest {
```
wli: SavePlanServletTest

wli (changes requested):

wli (approved):

PR #79, opened 2020-07-31
Download sections and meetings to datastore

savsa:
Previously, only courses were downloaded to datastore, but now all course, section, and meeting information is downloaded. 

Add JSONObject parsing for Sections and Meetings. Add tests.

### DIFF
--- a/src/main/java/com/google/collegeplanner/data/Meeting.java
+++ b/src/main/java/com/google/collegeplanner/data/Meeting.java
@@ -16,6 +16,9 @@ package com.google.collegeplanner.data;
 
 import java.text.ParseException;
 import java.time.DayOfWeek;
+import java.time.LocalTime;
+import java.time.format.DateTimeFormatter;
+import java.time.temporal.ChronoField;
 import java.util.ArrayList;
 import java.util.Collections;
 import org.json.simple.JSONArray;
@@ -42,6 +45,43 @@ public class Meeting {
     this.startTime = startTime;
     this.endTime = endTime;
 
+    assignDays(days);
+  }
+
+  public Meeting(JSONObject json) throws ParseException {
+    this.days = new ArrayList<DayOfWeek>();
+    this.startTime = parseTime((String) json.get("start_time"));
+    this.endTime = parseTime((String) json.get("end_time"));
+    this.room = (String) json.get("room");
+    this.building = (String) json.get("building");
+
+    assignDays((String) json.get("days"));
+  }
+
+  /**
+   * Convert 12-hour time to just minutes as an int.
+   * @param time The 12-hour time string eg. 12:30pm.
+   */
+  private int parseTime(String time) throws ParseException {
+    if (time == null || time == "") {
+      throw new ParseException("Invalid time format.", 0);
+    }
+    time = time.toUpperCase();
+
+    // Parse the 24-hour time to get the hours and minutes
+    LocalTime localTime = LocalTime.parse(time, DateTimeFormatter.ofPattern("h:mma"));
+    int hours = localTime.get(ChronoField.CLOCK_HOUR_OF_DAY);
+    int minutes = localTime.get(ChronoField.MINUTE_OF_HOUR);
+
+    // Convert the hours and minutes into just minutes
+    return 60 * hours + minutes;
+  }
+
+  /**
+   * Parses a string indicating which days have meetings and assigns the result to 'days'.
+   * @param days The string containing the days that have meetings.
+   */
+  private void assignDays(String days) throws ParseException {
     if (days.toUpperCase().contains("M")) {
       this.days.add(DayOfWeek.MONDAY);
     }
@@ -61,11 +101,6 @@ public class Meeting {
       // Since this is an academic scheduler, meetings cannot happen on weekends.
       throw new ParseException("Invalid Day, no Weekends on Academic Calendar", 0);
     }
-  }
-
-  public Meeting(JSONObject json) throws ParseException {
-    this((String) json.get("days"), (String) json.get("room"), (String) json.get("building"),
-        (int) json.get("start_time"), (int) json.get("end_time"));
   }
 
   public boolean conflictsWith(Meeting other) {
@@ -135,8 +170,18 @@ public class Meeting {
 
   public JSONObject toJSON() {
     JSONObject json = new JSONObject();
-    String daysString = "";
+    String daysString = getDaysAsString();
 
+    json.put("days", daysString);
+    json.put("room", room);
+    json.put("building", building);
+    json.put("start_time", startTime);
+    json.put("end_time", endTime);
+    return json;
+  }
+
+  public String getDaysAsString() {
+    String daysString = "";
     if (days.contains(DayOfWeek.MONDAY)) {
       daysString += "M";
     }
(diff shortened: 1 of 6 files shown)


ramyabuva (approved):

src/main/java/com/google/collegeplanner/servlets/DatastoreServlet.java
```diff
@@ -138,29 +141,52 @@ private void addCourses(JSONArray coursesArray, HttpServletResponse response) th
...
    * @param course The Course object.
    * @param courseEntity The Entity object that we want to add to datastore.
+   * @param response The HttpServletResponse object.
    */
-  private void addSectionsToCourse(JSONArray sectionsArray, Course course, Entity courseEntity) {
-    if (sectionsArray == null || course == null || courseEntity == null) {
+  private void addSectionsToCourse(JSONArray sectionsArray, Course course, Entity courseEntity,
+      HttpServletResponse response) throws IOException {
+    if (sectionsArray == null || course == null || courseEntity == null || response == null) {
+      respondWithError(HttpServletResponse.SC_INTERNAL_SERVER_ERROR, response);
```
wli: responding with an error and then returning here can cause some issues. The caller function doesn't really know that anything went wrong.

I think there's two ways to fix this:

1. Return a boolean from this function (true if it was successful), and then propagate that up to doPost, which can set the error properly.
2. Throw an exception and have doPost catch that exception, which then returns the error properly.

src/main/java/com/google/collegeplanner/servlets/DatastoreServlet.java
```diff
@@ -138,29 +141,52 @@ private void addCourses(JSONArray coursesArray, HttpServletResponse response) th
...
     ArrayList<EmbeddedEntity> sectionEntities = new ArrayList<EmbeddedEntity>();
     for (Object jsonObject : sectionsArray) {
-      // TODO(savsa): create Section class to handle JSON object parsing.
-      // For now, just store the section id.
       JSONObject sectionJson = (JSONObject) jsonObject;
+      Section section;
+      try {
+        section = new Section(sectionJson);
+      } catch (Exception e) {
+        respondWithError(HttpServletResponse.SC_INTERNAL_SERVER_ERROR, response);
```
wli: same here

src/main/java/com/google/collegeplanner/servlets/DatastoreServlet.java
```diff
@@ -179,4 +205,27 @@ private void addSectionsToCourse(JSONArray sectionsArray, Course course, Entity
...
+
+  /**
+   * Converts Meeting objects into Meeting entities and then adds them to the meetingEntities array.
+   * @param meetingEntities The ArrayList of Meeting entities.
+   * @param meetings The Array of Meeting objects.
+   */
+  private void addToMeetingEntities(ArrayList<EmbeddedEntity> meetingEntities, Meeting[] meetings,
+      HttpServletResponse response) throws IOException {
+    if (meetingEntities == null || meetings == null) {
+      respondWithError(HttpServletResponse.SC_INTERNAL_SERVER_ERROR, response);
```
wli: and here

wli (changes requested):

src/main/java/com/google/collegeplanner/servlets/DatastoreServlet.java
savsa: I'm actually thinking about throwing an error if one of the arguments is null, but then just `continue`-ing for parse exceptions. I don't think the whole program should stop because of a parse exception on just one course. I think it should still add the rest of the courses that aren't messed up.

wli: Yep, that makes sense. In that case, you probably would return a 200 then instead of returning 500 whenever any section fails. What you can do is also report the status in the response (X number of courses imported, Y failed, etc).

savsa: After `continue`-ing it should eventually reach the statement where I return status 200 in `doPost()`. Do you think returning how many succeeded is necessary?

src/main/java/com/google/collegeplanner/data/Meeting.java
```diff
@@ -42,6 +45,43 @@ public Meeting(String days, String room, String building, int startTime, int end
...
+   * @param time The 12-hour time string eg. 12:30pm.
+   */
+  private int parseTime(String time) throws ParseException {
+    if (time == null || time == "") {
+      throw new ParseException("Invalid time format.", 0);
+    }
+    time = time.toUpperCase();
+
+    // Parse the 24-hour time to get the hours and minutes
+    LocalTime localTime = LocalTime.parse(time, DateTimeFormatter.ofPattern("h:mma"));
```
triplehair: Curious what if `time` is not in the format of "h::mma", what will we get?

src/main/java/com/google/collegeplanner/servlets/DatastoreServlet.java
wli: How will you know if anything failed?

src/main/java/com/google/collegeplanner/servlets/DatastoreServlet.java
```diff
@@ -134,33 +141,52 @@ private void addCourses(JSONArray coursesArray, HttpServletResponse response) th
...
     // Loop through each course's sections.
     ArrayList<EmbeddedEntity> sectionEntities = new ArrayList<EmbeddedEntity>();
     for (Object jsonObject : sectionsArray) {
-      // TODO(savsa): create Section class to handle JSON object parsing.
-      // For now, just store the section id.
       JSONObject sectionJson = (JSONObject) jsonObject;
+      Section section;
+      try {
+        section = new Section(sectionJson);
+      } catch (Exception e) {
```
wli: please catch specific exceptions

src/main/java/com/google/collegeplanner/servlets/DatastoreServlet.java
```diff
@@ -87,18 +90,23 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
         // Successful. The default status code response is 200.
         return;
       }
-      addCourses(coursesArray, response);
+
+      try {
+        addCourses(coursesArray);
+      } catch (IllegalArgumentException e) {
+        respondWithError(HttpServletResponse.SC_INTERNAL_SERVER_ERROR, response);
```
wli: should this also `continue` instead of error...? i feel like the same logic from your other comment would apply here.

wli (changes requested):

src/main/java/com/google/collegeplanner/servlets/DatastoreServlet.java
```diff
@@ -87,18 +91,22 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
       }
-      addCourses(coursesArray, response);
+
+      try {
+        addCourses(coursesArray);
+      } catch (IllegalArgumentException e) {
+        continue;
+      }
     } while (page++ < PAGE_LIMIT);
   }
```
wli, line 101: i just noticed that we don't return...anything? No `application/json`, no response message. Something to fix up in a separate PR.

wli (approved):

src/main/java/com/google/collegeplanner/servlets/DatastoreServlet.java
savsa: I'll create an issue.

triplehair: +1

triplehair (approved):

src/main/java/com/google/collegeplanner/data/Meeting.java
savsa: `DateTimeParseException`. Just added an error check.

PR #80, opened 2020-07-31
Create page for saved planners and add util.js

ramyabuva:
I created util.js to hold shared functions to create alerts and planner tables that were necessary for more than one page. 

https://screenshot.googleplex.com/TYeepiUnCCu

### DIFF
--- a/src/main/webapp/js/lib/login.js
+++ b/src/main/webapp/js/lib/login.js
@@ -12,6 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+/**
+ * Array of functions to execute after signing in in addition to onSignIn
+ * @type {!Array<function>}
+ */
 const postSignInHandlers = [];
 
 /**
--- a/src/main/webapp/js/lib/login.js
+++ b/src/main/webapp/js/lib/login.js
@@ -12,13 +12,15 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+const postSignInHandlers = [];
+
 /**
  * Gets user information and populates the topbar with their image and name
  * @param {GoogleUser} googleUser the currently logged in user
  */
 function onSignIn(googleUser) {
   const profile = googleUser.getBasicProfile();
-  // Add user name and photo to page
+  // Add user name and photo to page.
   const userInfo = document.getElementById('user-info');
   const userName = document.createElement('span');
   userName.setAttribute('class', 'mr-2 d-none d-lg-inline text-gray-600 small');
@@ -29,8 +31,10 @@ function onSignIn(googleUser) {
   userInfo.innerText = ''
   userInfo.appendChild(userName);
   userInfo.appendChild(userPhoto);
-  // make dropdown to sign out visible
+  // Make dropdown to sign out visible.
   document.getElementById('user-dropdown').hidden = false;
+  // Execute each postSignInHandler
+  postSignInHandlers.forEach(func => func(googleUser));
 }
 
 /**
@@ -43,10 +47,18 @@ function signOut() {
   });
 }
 
+function registerPostSignInHandler(handler) {
+  postSignInHandlers.push(handler);
+}
+
 /**
  * Set window onSignIn so the function is executed on login.
  * Exported so that the function can be used on all pages of site.
  */
 window.onSignIn = onSignIn;
 
-export default {onSignIn: onSignIn, signOut: signOut};
+export default {
+  onSignIn: onSignIn,
+  signOut: signOut,
+  registerPostSignInHandler: registerPostSignInHandler
+};
--- a/src/main/webapp/js/lib/utils.js
+++ b/src/main/webapp/js/lib/utils.js
@@ -1,0 +1,77 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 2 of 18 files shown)


src/main/webapp/js/lib/utils.js
```diff
@@ -0,0 +1,77 @@
...
+ */
+function createAlert(message, type, container) {
+  const alert = document.createElement('div');
+  alert.setAttribute('class', `alert alert-${type}`);
+  alert.setAttribute('role', 'alert');
+  alert.appendChild(document.createTextNode(message));
+  container.appendChild(alert);
+}
+
+export default {
```
wli: I just learned that default exports are banned by style guide (https://google.github.io/styleguide/jsguide.html#named-vs-default-exports)

src/main/webapp/js/pages/index.js
```diff
@@ -12,9 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+import Auth from '../lib/login.js';
```
wli: Did a linter re-sort this? If not, I believe that typically sort order is by file name, not by the import name (since you can import lots of things out of it).

src/main/webapp/js/lib/utils.js
```diff
@@ -0,0 +1,77 @@
...
+  courseContainer.appendChild(spinner);
+}
+
+/**
+ * Creates table from a 2D array
+ * @param {Object} tableData 2D array with separation of courses
+ * @param {Object} creditsData Array with credits for each semester
+ * @param {Element} courseContainer container for course list
+ */
+function createTable(tableData, creditsData, courseContainer) {
```
wli: The name seems more generic than it should be. It only creates one type of table. Are you sure this belongs in this util file? If so, you might want to rename the function name to be clearer too.

src/main/webapp/js/pages/savedPlanners.js
```diff
@@ -0,0 +1,90 @@
...
+  card.appendChild(body);
+  document.getElementById('plan-column').appendChild(card);
+}
+
+window.addEventListener('load', () => {
+  function onSignIn(googleuser) {
+    Auth.onSignIn(googleuser);
+    getSavedPlans();
+  }
+  window.onSignIn = onSignIn;
```
wli: can we modify the Auth library to take in an additional callback, instead of overwriting a global here?

src/main/webapp/vendor/startbootstrap/saved-planner-content.jsp
```diff
@@ -0,0 +1,15 @@
...
+<%@ taglib uri="http://java.sun.com/jsp/jstl/fmt" prefix ="fmt" %>
+<!-- Begin Page Content -->
+<div class="container-fluid">
+  <!-- Page Heading -->
+  <div class="d-sm-flex align-items-center justify-content-between mb-4" id="alert-container">
+    <h1 class="h3 mb-0 text-gray-800">Planner: My Saved Plans</h1>
+  </div>
+  <div class="row">
+    <!-- Course Selection Area -->
+    <div class="col-xl-12 col-lg-7" id= "plan-column">
```
wli: are the other 5 width units taken up by the sidebar in `lg-7`? What's the rationale here?

wli (changes requested):

src/main/webapp/vendor/startbootstrap/saved-planner-content.jsp
ramyabuva: I'm actually not sure, it seems to look the same when i remove it from the class -- I was following the pattern set in the UI templates, but I think i'll just remove it.

wli: Ah ok. Normally this changes based on your screen width (if you resize your window, the elements' widths and positions may shift around). Probably a bit beyond the scope of the work here, but it's good to learn someday.

src/main/webapp/js/lib/utils.js
ramyabuva: I did it for utils.js in this PR, but I'll open another one for the rest of the files.

src/main/webapp/js/pages/savedPlanners.js
ramyabuva: I'm not sure how to do this without modifying the functionality for other pages.

savsa (approved):

src/main/webapp/js/pages/savedPlanners.js
wli: For example, in the Auth library, if you add:

```
const postSignInHandlers = [];

function registerPostSignInHandler(handler) {
  postSignInHandlers.push(handler);
}
```

you could call it with:

```
Auth.registerPostSignInHandler(() => {
  getSavedPlans();
});
```

You can call this immediately, without listening to the `load` event here too.

wli (changes requested):

src/main/webapp/js/pages/savedPlanners.js
ramyabuva: Ah i see, that's smart

src/main/webapp/js/lib/login.js
```diff
@@ -12,13 +12,15 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+const postSignInHandlers = [];
```
wli, line 19: add type information and documentation for this

wli (approved):

PR #82, opened 2020-07-31
Remove default imports from js files

ramyabuva:


src/main/webapp/js/lib/login.js
```diff
@@ -49,4 +49,7 @@ function signOut() {
  */
 window.onSignIn = onSignIn;
 
-export default {onSignIn: onSignIn, signOut: signOut};
+export const Auth = {
```
wli, line 64: i feel like this should be `auth.js` then?

savsa (approved):

src/main/webapp/js/lib/courseSelector.js
```diff
@@ -168,7 +168,7 @@ function getCourseInfo() {
   return courseInfo;
 }
 
-export default {
+export const CollegePlanner = {
```
wli: shoudl this be CourseSelector instead of CollegePlanner?

wli (changes requested):

wli (approved):

triplehair (approved):

PR #83, opened 2020-07-31
Add option to delete saved plans

ramyabuva:


src/main/java/com/google/collegeplanner/servlets/DeletePlanServlet.java
```diff
@@ -0,0 +1,38 @@
+// Copyright 2019 Google LLC
```
wli: just noticed - 2019 -> 2020?

src/main/java/com/google/collegeplanner/servlets/DeletePlanServlet.java
```diff
@@ -0,0 +1,38 @@
...
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/** Servlet responsible for deleting plans. */
+@WebServlet("/delete-plan")
+public class DeletePlanServlet extends HttpServlet {
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    long id = Long.parseLong(request.getParameter("id"));
```
wli: do we need to catch and respondWithError for null parameter and also if the id is missing?

src/main/webapp/js/pages/savedPlanners.js
```diff
@@ -56,6 +57,23 @@ function createCard(planName, tableData, creditsData) {
   title.setAttribute('class', 'm-0 font-weight-bold text-primary');
   title.innerText = planName;
   header.appendChild(title);
+  // Create delete button.
+  const deleteButtonElement = document.createElement('button');
+  deleteButtonElement.innerHTML =
+      '<i class="fas fa-trash-alt" id="modeicon"></i>';
+  deleteButtonElement.setAttribute(
+      'class', 'rounded-circle border-0 btn-lg float:right');
```
wli: are you sure `float:right` is a class? Should this be a separate CSS style declaration?

src/main/webapp/js/pages/savedPlanners.js
```diff
@@ -66,6 +84,16 @@ function createCard(planName, tableData, creditsData) {
...
 }
 
+/**
+ * Tells the server to delete the plan.
+ * @param {Number} id The id of the plan to delete.
+ */
+function deletePlan(id) {
+  const params = new URLSearchParams();
+  params.append('id', id);
+  fetch('/delete-plan', {method: 'POST', body: params});
```
wli: no user feedback if the fetch fails here. you probably want to try/catch.

src/test/java/com/google/collegeplanner/DeletePlanServletTest.java
```diff
@@ -0,0 +1,97 @@
...
+    // Put 2 entities in datastore. 
+    Entity planEntity1 = new Entity("Plan");
+    datastore.put(planEntity1);
+
+    Entity planEntity2 = new Entity("Plan");
+    datastore.put(planEntity2);
+
+    // Delete an element and check that there are still 2 elements in datastore.
+    when(request.getParameter("id")).thenReturn("3");
+    servlet.doPost(request, null);
```
wli: you should check the HTTP status of the output here too. it shouldn't be a 500.

src/test/java/com/google/collegeplanner/DeletePlanServletTest.java
```diff
@@ -0,0 +1,97 @@
...
+    // Put 2 entities in datastore. 
+    Entity planEntity1 = new Entity("Plan");
+    datastore.put(planEntity1);
+
+    Entity planEntity2 = new Entity("Plan");
+    datastore.put(planEntity2);
+
+    // Delete an element and check that there is only 1 element in datastore.
+    when(request.getParameter("id")).thenReturn("1");
+    servlet.doPost(request, null);
```
wli: check HTTP status here

wli (changes requested):

src/test/java/com/google/collegeplanner/DeletePlanServletTest.java
```diff
@@ -0,0 +1,133 @@
...
+    // Delete an element and check that there are still 2 elements in datastore.
+    when(request.getParameter("id")).thenReturn(null);
+    servlet.doPost(request, response);
+
+    verify(response, times(1)).setStatus(HttpServletResponse.SC_BAD_REQUEST);
+    verify(response, never()).setStatus(not(eq(HttpServletResponse.SC_BAD_REQUEST)));
+    String expectedJson =
+        "{\"message\":\"Invalid id for datastore deletion.\",\"status\":\"error\"}";
+    JSONAssert.assertEquals(expectedJson, stringWriter.toString(), JSONCompareMode.STRICT);
+  }
```
naaoli, line 137: Maybe I missed it but will the delete button be shown when there are no plans?  Because if i's shown you'll need a test for testing out the delete button when there are no saved plans

ramyabuva: The delete buttons are only created along with the cards for each schedule

savsa (approved):

src/main/java/com/google/collegeplanner/servlets/DeletePlanServlet.java
```diff
@@ -0,0 +1,44 @@
...
+@WebServlet("/delete-plan")
+public class DeletePlanServlet extends BaseServlet {
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    try {
+      long id = Long.parseLong(request.getParameter("id"));
+
+      Key planEntityKey = KeyFactory.createKey("Plan", id);
+      DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+      datastore.delete(planEntityKey);
```
wli, line 40: you should output some kind of json response when it is successful

src/main/java/com/google/collegeplanner/servlets/DeletePlanServlet.java
```diff
@@ -0,0 +1,44 @@
...
+import com.google.appengine.api.datastore.KeyFactory;
+import com.google.collegeplanner.servlets.BaseServlet;
+import java.io.IOException;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/** Servlet responsible for deleting plans. */
+@WebServlet("/delete-plan")
```
wli: should this be `/api/delete-plan`?

wli (changes requested):

src/main/webapp/js/pages/savedPlanners.js
```diff
@@ -66,6 +84,22 @@ function createCard(planName, tableData, creditsData) {
...
 
+/**
+ * Tells the server to delete the plan.
+ * @param {Number} id The id of the plan to delete.
+ */
+function deletePlan(id) {
+  const params = new URLSearchParams();
+  params.append('id', id);
+  try {
+    fetch('/delete-plan', {method: 'POST', body: params});
```
wli: you'll want to check if the response is `.ok`, and put up an error if it is not.

wli (changes requested):

src/main/webapp/js/pages/savedPlanners.js
```diff
@@ -66,6 +83,31 @@ function createCard(planName, tableData, creditsData) {
...
+  const params = new URLSearchParams();
+  params.append('id', id);
+  let response;
+  try {
+    response =
+        await fetch('/api/planner/delete', {method: 'POST', body: params});
+  } catch (err) {
+    Util.createAlert(
+        'Could not delete element.', 'danger',
+        document.getElementById('plan-name'));
```
wli, line 100: I think you want to return here instead of falling through. Otherwise `response` will be undefined and you will probably hit an exception on line 102 when attempting access `undefined.ok`.

wli (changes requested):

src/main/webapp/js/pages/savedPlanners.js
ramyabuva: i totally missed that, thanks

wli (approved):
Feel free to merge.

PR #85, opened 2020-08-01
Link up SemesterScheduler.java to webpage

naaoli:
This PR connects the SemesterScheduler.java backend code with the webpage with a servlet and some modifications to the index.js file along with some modifications to other code.


src/main/webapp/js/pages/index.js
```diff
@@ -37,18 +37,48 @@ function addScheduleToCalendar(schedule, courseInfo, selected) {
...
   // Hard code return from algorithm servlet.
+  /*
   const sections1 = {};
   selected.forEach(
       course => {sections1[course] = courseInfo[course].sections[0]});
   const sections2 = {};
   selected.forEach(
       course => {sections2[course] = courseInfo[course].sections[1]});
-  const schedules = [sections1, sections2];
+  const schedules = [sections1, sections2];*/
```
ramyabuva: You can get rid of this code completely

src/main/webapp/js/pages/index.js
```diff
@@ -37,18 +37,48 @@ function addScheduleToCalendar(schedule, courseInfo, selected) {
...
+      body: JSON.stringify(data),
+    });
+    scheduleObject = await response.json();
+    if (response.ok) {
+      return scheduleObject.schedules;
+    } else {
+      return null;
+    }
+  } catch (err) {
+    CollegePlanner.createAlert('An error occurred', 'danger', courseContainer);
```
ramyabuva: You should merge with master, this will have to change to CourseSelector.createAlert

src/main/java/com/google/collegeplanner/servlets/SchedulerServlet.java
```diff
@@ -0,0 +1,143 @@
...
+    json.put("schedules", schedulesJson);
+
+    return json;
+  }
+
+  /**
+   * Converts a given JSONArray of sections into an ArrayList of
+   * sections.
+   */
+  private ArrayList<Section> convertJSONArraytoArrayList(JSONArray json) {
```
ramyabuva: Maybe make the function name something more specific to the function of converting sections, and document the parameters

src/main/java/com/google/collegeplanner/servlets/SchedulerServlet.java
```diff
@@ -0,0 +1,143 @@
...
+  /**
+   * Gets the JSON Representation of the body of the POST request
+   */
+  private JSONObject getBody(HttpServletRequest request)
+      throws IOException, ParseException, NullPointerException {
+    String strBody =
+        request.getReader().lines().collect(Collectors.joining(System.lineSeparator()));
+    JSONParser parser = new JSONParser();
+    return (JSONObject) parser.parse(strBody);
+  }
```
ramyabuva: This function is in BaseServlet, so you don't need a repeat function definition

ramyabuva (changes requested):

src/main/java/com/google/collegeplanner/data/Meeting.java
```diff
@@ -209,4 +221,23 @@ public String getDaysAsString() {
...
+    int minutes = time % 60;
+    if (hours > 12) {
+      hours -= 12;
+    } else if (hours == 0) {
+      hours = 12;
+    }
+    if (hours >= 12) {
+      amPm = "pm";
+    }
+    return (Integer.toString(hours) + ":" + Integer.toString(minutes) + amPm);
```
wli: instead of 05:06pm, this would show 5:6pm right? Is that expected?

src/main/java/com/google/collegeplanner/servlets/SchedulerServlet.java
```diff
@@ -0,0 +1,138 @@
...
+    JSONArray jsonArray;
+    ApiUtil apiUtil = new ApiUtil();
+
+    courseList = new ArrayList<String>();
+    courses = new ArrayList<ArrayList<Section>>();
+    for (Object obj : classes) {
+      courseId = (String) obj;
+      courseList.add(courseId);
+      try {
+        uri = new URI("https://api.umd.io/v1/courses/" + courseId + "/sections");
```
wli: courseId should probably be escaped for safety

src/main/java/com/google/collegeplanner/servlets/SchedulerServlet.java
```diff
@@ -0,0 +1,138 @@
...
+      } catch (URISyntaxException e) {
+        respondWithError(HttpServletResponse.SC_INTERNAL_SERVER_ERROR, response);
+        return;
+      }
+
+      jsonArray = apiUtil.getJsonArray(uri);
+      if (jsonArray == null) {
+        respondWithError(HttpServletResponse.SC_INTERNAL_SERVER_ERROR, response);
```
wli, line 98: instead of `respondWithError` here, this function should either throw an exception, or otherwise return an error. Otherwise, your HTTP response will be an error, but you will still return a blank/undefined schedules object printed out on line 62.

src/main/java/com/google/collegeplanner/servlets/SchedulerServlet.java
```diff
@@ -0,0 +1,138 @@
...
+   */
+  private JSONObject getSchedules() {
+    JSONObject json = new JSONObject();
+    JSONObject temp;
+    JSONArray schedulesJson = new JSONArray();
+    SemesterScheduler scheduler = new SemesterScheduler(courses);
+    ArrayList<Schedule> possibleSchedules = scheduler.getPossibleSchedules();
+
+    for (Schedule schedule : possibleSchedules) {
+      temp = schedule.toJSON();
```
wli: no need to define temp, you can replace this with:

`schedulesJson.add(schedule.toJSON());`

src/main/java/com/google/collegeplanner/servlets/SchedulerServlet.java
```diff
@@ -0,0 +1,138 @@
...
+    return json;
+  }
+
+  /**
+   * Converts a given JSONArray of sections into an ArrayList of
+   * sections.
+   * @param json The JSONArray of Sections that will be converted into an
+   * ArrayList of Sections
+   */
+  private ArrayList<Section> convertSectionJSONToSectionArrayList(JSONArray json) {
```
wli: naming conventions dictate `convertSectionJSONToSectionArrayList` -> `convertSectionJsonToSectionArrayList`

wli (changes requested):

src/main/java/com/google/collegeplanner/servlets/SchedulerServlet.java
naaoli: what do you mean by escaped for safety?

wli: @savsa @ramyabuva there's one other java example for uri-escaping this right? can you point naaoli to it?

src/main/java/com/google/collegeplanner/data/Meeting.java
```diff
@@ -205,8 +218,32 @@ public String getDaysAsString() {
...
+    // and returns a string of the number represented in two digits. 
+    // Examples: 1 -> "01"      12 -> "12"
+    DecimalFormat formatter = new DecimalFormat("00");
+    if (hours > 12) {
+      hours -= 12;
+    } else if (hours == 0) {
+      hours = 12;
+    }
+    if (hours >= 12) {
+      amPm = "pm";
```
triplehair, line 271: If the time is 10 mins after midnight, will it be changed to "pm" instead?

src/main/java/com/google/collegeplanner/servlets/SchedulerServlet.java
```diff
@@ -0,0 +1,136 @@
...
+    try {
+      body = getPostRequestBody(request);
+      selectedClasses = (JSONArray) body.get("selectedClasses");
+    } catch (NumberFormatException | ClassCastException | ParseException | NullPointerException e) {
+      respondWithError(
+          "Invalid body for POST request.", HttpServletResponse.SC_BAD_REQUEST, response);
+      return;
+    }
+
+    prepareLists(selectedClasses, response);
```
triplehair: Should we return here if prepareLists return an exception?

src/main/java/com/google/collegeplanner/servlets/SchedulerServlet.java
ramyabuva: I think the URI class automatically uri-escapes the input string, the only time we uri-escaped was in the Javascript

ramyabuva (approved):

src/main/java/com/google/collegeplanner/servlets/SchedulerServlet.java
naaoli: I think that is a good idea, will be doing that in the next commit

src/main/java/com/google/collegeplanner/servlets/SchedulerServlet.java
wli: Hmm, i'm not sure that's right. In this case, if `courseId` was `"foo/bar?baz="`, it could modify the url quite a bit, as it's just a raw string when you pass it in.

src/main/java/com/google/collegeplanner/data/Meeting.java
naaoli: I messed up the logic slightly, will fix it

src/main/webapp/js/pages/index.js
```diff
@@ -37,18 +37,42 @@ function addScheduleToCalendar(schedule, courseInfo, selected) {
...
+    }
+  } catch (err) {
+    CourseSelector.createAlert('An error occurred', 'danger', courseContainer);
+    return null;
+  }
+}
+
+document.querySelector('.course-list').addEventListener('click', async () => {
+  const selected = CollegePlanner.getSelected();
+  const courseInfo = CollegePlanner.getCourseInfo();
```
ramyabuva: These also have to be changed to CourseSelector -- whenever you merge with master, make sure you test things on the devserver to make sure all of the modules still work.

src/main/java/com/google/collegeplanner/servlets/SchedulerServlet.java
```diff
@@ -0,0 +1,140 @@
...
+    courseList = new ArrayList<String>();
+    courses = new ArrayList<ArrayList<Section>>();
+    for (Object obj : classes) {
+      courseId = (String) obj;
+      courseList.add(courseId);
+      try {
+        uri = new URI("https://api.umd.io/v1/courses/" + courseId + "/sections");
+      } catch (URISyntaxException e) {
+        respondWithError(HttpServletResponse.SC_INTERNAL_SERVER_ERROR, response);
+        throw new IOException(courseId + " cannot be found in the umd.io database");
```
wli: this isn't really an IOException. What if you just let the URISyntaxException bubble up to the doPost method?

src/main/java/com/google/collegeplanner/servlets/SchedulerServlet.java
wli: This isn't quite right still. Since you don't throw any error, it would continue execution into the "success" case.

src/main/java/com/google/collegeplanner/servlets/SchedulerServlet.java
```diff
@@ -0,0 +1,140 @@
...
+   * sections.
+   * @param json The JSONArray of Sections that will be converted into an
+   * ArrayList of Sections
+   */
+  private ArrayList<Section> convertSectionJsonToSectionArrayList(JSONArray json) {
+    ArrayList<Section> list = new ArrayList<Section>();
+    for (Object obj : json) {
+      try {
+        list.add(new Section((JSONObject) obj));
+      } catch (Exception e) {
```
wli: We should catch a narrower exception. Please see the following excerpt from the Java readability guidelines:

```
Does the underlying code actually throw Exception, or a derived type of exception? You should generally catch the narrowest exception type.

Only catch Exception when you legitimately want to handle any exception that may arise, including unchecked ones you don't expect to actually happen.

Unchecked exceptions are usually used to convey the message that either the exception can be prevented by proper use of the API, or that callers should rarely catch this exception because there's nothing they can do about it. However, `catch (Exception e)` is fine if your intention is to gracefully downgrade the flow regardless (e.g., a catch-all in the main() method, which prevents any exceptions from being propagated to the Java VM).
```

src/main/webapp/js/pages/index.js
```diff
@@ -37,18 +37,42 @@ function addScheduleToCalendar(schedule, courseInfo, selected) {
...
+  try {
+    response = await fetch('/api/scheduler', {
+      method: 'POST',
+      body: JSON.stringify(data),
+    });
+    scheduleObject = await response.json();
+    if (response.ok) {
+      return scheduleObject.schedules;
+    } else {
+      return null;
```
wli, line 60: I think you need an alert here too. The difference here is that this case handles when the API returns an empty/bad response, whereas the `catch` below handles cases where it didn't even make it to your API.

wli (changes requested):

src/main/java/com/google/collegeplanner/servlets/SchedulerServlet.java
naaoli: I really can't make it any narrower as the exception being thrown is a java.text.ParseException. However, I already have org.json.simple.parser.ParseException being imported.

wli: To solve the duplicate name, you can import `java.text`, and reference `text.ParseException` -- or vice versa.

src/main/java/com/google/collegeplanner/servlets/SchedulerServlet.java
```diff
@@ -0,0 +1,141 @@
...
+        uri = new URI("https://api.umd.io/v1/courses/" + courseId + "/sections");
+      } catch (URISyntaxException e) {
+        respondWithError(HttpServletResponse.SC_INTERNAL_SERVER_ERROR, response);
+        return;
+      }
+
+      jsonArray = apiUtil.getJsonArray(uri);
+      if (jsonArray == null) {
+        respondWithError(HttpServletResponse.SC_INTERNAL_SERVER_ERROR, response);
+        throw new ParseException(0);
```
wli: throwing ParseException here unfortunately isn't quite right. You're not parsing an input as this point. Honestly, it might make sense to create an InvalidApiResponseError exception for you to use in this case. Alternatively, you can return a boolean in this function (true/false) and just return a generic exception for now if that helps.

savsa (approved):

triplehair (approved):

src/main/java/com/google/collegeplanner/servlets/SchedulerServlet.java
wli: https://github.com/googleinterns/step145-2020/blob/6485b74a3bfd38d8e4346a7b7f88bda00c253057/src/main/java/com/google/collegeplanner/servlets/DatastoreServlet.java#L141

src/main/java/com/google/collegeplanner/servlets/SchedulerServlet.java
```diff
@@ -0,0 +1,140 @@
...
+  /**
+   * This method loads the courses and courseList ArrayLists with correct
+   * information. courses is loaded with the lists of sections for each
+   * course while courseList is loaded with the courseID's of the selected
+   * courses. Returns true if lists were prepared correctly, false otherwise.
+   * @param classes The JSONArray of courseIds of the selected classes
+   * @param response The HttpServletResponse object
+   */
+  private boolean prepareLists(JSONArray classes, HttpServletResponse response) 
+  throws IOException {
```
wli: nit: i think this needs to be indented by another 4 spaces, but we can defer to the linter if this is what it thinks is correct.

wli (approved):

PR #87, opened 2020-08-03
Let CourseListServlet and SectionServlet pull data from datastore

### LINKED ISSUE #86: Have CourseListServlet and SectionServlet pull from datastore

savsa:
- CourseListServlet and Section now pull data directly from datastore and not the UMD API.

- Gson can directly serialize classes. `@SerializedName` was added to control what string each class variable gets serialized to when converting it into json.

- `Course`, `Section`, and `Meeting` classes were given constructors to convert datastore `Entity`/`EmbeddedEntity`s into their respective Course/Section/Meeting objects, which are then serialized into json.

- Add tests to make sure we're still getting correct output using datastore.

- Add tests to make sure the correct error codes are sent on error.

Closes #86.

### DIFF
--- a/src/main/java/com/google/collegeplanner/data/Course.java
+++ b/src/main/java/com/google/collegeplanner/data/Course.java
@@ -14,6 +14,9 @@
 
 package com.google.collegeplanner.data;
 
+import com.google.appengine.api.datastore.EmbeddedEntity;
+import com.google.appengine.api.datastore.Entity;
+import com.google.gson.annotations.SerializedName;
 import java.text.ParseException;
 import java.util.ArrayList;
 import java.util.GregorianCalendar;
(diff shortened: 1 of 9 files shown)


ramyabuva (approved):

wli (approved):

src/main/java/com/google/collegeplanner/data/Course.java
```diff
@@ -129,6 +137,20 @@ public Course(JSONObject json) throws ParseException {
...
+        ((Long) courseEntity.getProperty("credits")).intValue(),
+        (String) courseEntity.getProperty("dept_id"),
+        (String) courseEntity.getProperty("description"),
+        (String) courseEntity.getProperty("coreqs"), (String) courseEntity.getProperty("prereqs"),
+        (String) courseEntity.getProperty("restrictions"),
+        (String) courseEntity.getProperty("additional_info"),
+        (String) courseEntity.getProperty("credit_granted_for"));
+  }
+
+  public Course(Entity courseEntity, ArrayList<EmbeddedEntity> sections) {}
```
triplehair: Is this method being used anywhere?

triplehair (approved):

src/main/java/com/google/collegeplanner/data/Course.java
savsa: Nope, removed.

PR #89, opened 2020-08-03
Add minor UI fixes 

ramyabuva:
This PR fixes the table rendering for plan, and moves the calendar so that adding additional courses does not move it down.


savsa (approved):

src/main/webapp/js/lib/util.js
```diff
@@ -55,6 +56,21 @@ function createPlanTable(tableData, creditsData, courseContainer) {
   courseContainer.appendChild(table);
 }
 
+/**
+ * Rectangularizes matrix by ensuring each row has an equal number of elements.
+ * @param {Object} data The 2d array to rectangularize
+ */
+function rectangularize2dMatrix(data) {
+  const longestRow =
```
wli: why can't this be:

`const longestRow = Math.max(data.map(arr => arr.length));`

wli (changes requested):

src/main/webapp/js/lib/util.js
ramyabuva: I tried that first, it results in NaN for some reason.

wli: Ah, you need: `Math.max(...data.map(arr => arr.length));`

Math.max takes in a variable number of arguments, not an array.

wli (changes requested):

src/main/webapp/js/lib/util.js
ramyabuva: that works!

src/main/webapp/js/pages/planner.js
```diff
@@ -92,6 +92,11 @@ async function savePlan(courseList) {
       method: 'POST',
       body: JSON.stringify(data),
     });
+    document.getElementById('save-button-text').innerText = 'Saved!';
+    document.getElementById('save-plan-prompt')
+        .classList.remove('btn-secondary');
+    document.getElementById('save-plan-prompt').classList.add('btn-primary');
```
wli, line 98: question - is the user allowed to save again? Should this button be disabled instead?

ramyabuva: I decided  not to disable it so they could save it again under a different name if they wanted.

wli: sounds good.

wli (approved):

triplehair (approved):

